PR #17, opened 2023-08-28
feat: add Dropdown Menu component

fdemb:
Dropdown Menu component pasted from the original and modified for Vue, with names and slots, using JSX.

PR #18, opened 2023-08-28
feat: add Alert and Avatar components

fdemb:
Adds misc components: Alert and Avatar.


fdemb:
If someone is able to fix TS errors on component HTML attributes- feel free to do that!

fdemb:
Same thing as #17.

PR #37, opened 2023-09-08
style: prevent linked whitespace in footer

vintagesucks:
An alternative solution would be to leave the texts on one line in the source. However, I did not want to change the structure of your code too much here, hence `inline-block`. I also removed the superfluous `text-sm` classes, as this size is already set in the parent element.

#### Before
before

#### After
after


zernonia (approved):
lgtm! Thanks for the fix! 😁

PR #38, opened 2023-09-09
fix(cli): break caused by non-existence of `tsconfig.app.json`

Dunqing:
The project I created through `vite` does not have `tsconfig.app.json`, So we should handle its non-existent situation.


Hyrran (approved):

k11q (approved):
lgtm! Thanks!

PR #42, opened 2023-09-10
refactor(apps/www): use `vite` postcss depdendecy, use `tsx` instead of `ts-node`

sadeghbarati:
- Vite already has [postcss](https://github.com/vitejs/vite/blob/main/packages/vite/package.json#L74) dependency so,
just to have clear workspace, we can remove `postcss.config.js` and move it to the Vite config file `(.vitepress/config.mts)`

- Use `tsx` instead of `ts-node`

- Fix excessive stack depth comparing types in VitePress `(config.mts)` by updating Vite

---

<details>
  <summary>
DropdownAction in DataTableDemo.vue caused scroll-x on the body because of `sr-only` class, added a relative wrapper to it to prevent scroll-x
  
  </summary>

image

</details>


sadeghbarati:
Before we add a dependency management tool like Renovate, I would suggest using [Taze](https://github.com/antfu/taze) to keep dev|dependencies updated

zernonia:
Thanks for the PR @sadeghbarati ! it looks good! Could you help setup Taze as per your suggestion? 😁

zernonia (approved):

PR #43, opened 2023-09-10
fix(cli): shouldn't load tsconfig when typescript disabled

Dunqing:
BTW, The javascript support doesn't look good,

The `SFC` file is still based `typescript` 
image


zernonia:
Thanks @Dunqing !

Yup I've excluded the SFC transformation, thus some `relative import` and `SFC` doesn't transform correctly. Your suggestion here should fix the JS issue, however not the transformation.

Let's handle that in another PR ya!

PR #44, opened 2023-09-10
feat: add profile section in forms example

ahmedmayara:
I did not manage to make the Form component using VeeValidate and Zod but I made the profile section in the forms example work using Zod, I will try to complete the rest of the pages later today.


zernonia (approved):
lgtm!

PR #45, opened 2023-09-11
feat(cli): support remove type when non-typescript

Dunqing:
Currently has bug needs to be fixed in https://github.com/cyco130/detype/pull/9


zernonia:
@Dunqing  any win with this? 😁

Dunqing:
> @Dunqing any win with this? 😁

Still has a blocking bug that needs to wait for `detype` to fix, or is there another better remove types related library?

sadeghbarati:
@Dunqing 

I tested your branch, it's removing types but `index.ts` remains as ts file

maybe mkdist source could help you [`vueLoader`](https://github.com/unjs/mkdist/blob/main/src/loaders/vue.ts) - [`jsLoader`](https://github.com/unjs/mkdist/blob/main/src/loaders/js.ts)


> [!NOTE]  
>  `vueLoader` source code [excluding script setup syntax](https://github.com/unjs/mkdist/blob/main/src/loaders/vue.ts#L116) from detype proccessing

- `detype` accept a string
- `mkdist` accept a directory

Dunqing:
> @Dunqing
> 
> I tested your branch, it's removing types but `index.ts` remains as ts file
> 
> maybe mkdist source could help you [`vueLoader`](https://github.com/unjs/mkdist/blob/main/src/loaders/vue.ts) - [`jsLoader`](https://github.com/unjs/mkdist/blob/main/src/loaders/js.ts)
> 
> Note `vueLoader` source code [excluding script setup syntax](https://github.com/unjs/mkdist/blob/main/src/loaders/vue.ts#L116) from detype proccessing
> 
> * `detype` accept a string
> * `mkdist` accept a directory

Thank you for these! These loaders look good, However, if it's going to be re-implemented based on the loader, I'd prefer to patch and fix bugs for `detype` for the time being.

Dunqing:
I think it's ready for review now.

zernonia:
@sadeghbarati  could you help me test this branch again? 😁

sadeghbarati:
Yeah sure

sadeghbarati:
@Dunqing 👋 

last time I tested your branch with `pnpm start init/add` directory was like this

```
📦cli
 ┣ 📂components
 ┃ ┗ 📂ui
 ┃ ┃ ┗ 📂accordion
 ┃ ┃ ┃ ┣ 📜Accordion.vue
 ┃ ┃ ┃ ┣ 📜AccordionContent.vue
 ┃ ┃ ┃ ┣ 📜AccordionItem.vue
 ┃ ┃ ┃ ┣ 📜AccordionTrigger.vue
 ┃ ┃ ┃ ┗ 📜index.ts
 ┣ 📂lib
 ┃ ┗ 📜utils.js
```

now it's look like this
```
📦cli
 ┣ 📂@
 ┃ ┣ 📂components
 ┃ ┃ ┗ 📂ui
 ┃ ┃ ┃ ┗ 📂accordion
 ┃ ┃ ┃ ┃ ┣ 📜Accordion.vue
 ┃ ┃ ┃ ┃ ┣ 📜AccordionContent.vue
 ┃ ┃ ┃ ┃ ┣ 📜AccordionItem.vue
 ┃ ┃ ┃ ┃ ┣ 📜AccordionTrigger.vue
 ┃ ┃ ┃ ┃ ┗ 📜index.js
 ┃ ┗ 📂lib
 ┃ ┃ ┗ 📜utils.js
```

Dunqing:
> @Dunqing 👋
> 
> last time I tested your branch with `pnpm start init/add` directory was like this
> 
> ```
> 📦cli
>  ┣ 📂components
>  ┃ ┗ 📂ui
>  ┃ ┃ ┗ 📂accordion
>  ┃ ┃ ┃ ┣ 📜Accordion.vue
>  ┃ ┃ ┃ ┣ 📜AccordionContent.vue
>  ┃ ┃ ┃ ┣ 📜AccordionItem.vue
>  ┃ ┃ ┃ ┣ 📜AccordionTrigger.vue
>  ┃ ┃ ┃ ┗ 📜index.ts
>  ┣ 📂lib
>  ┃ ┗ 📜utils.js
> ```
> 
> now it's look like this
> 
> ```
> 📦cli
>  ┣ 📂@
>  ┃ ┣ 📂components
>  ┃ ┃ ┗ 📂ui
>  ┃ ┃ ┃ ┗ 📂accordion
>  ┃ ┃ ┃ ┃ ┣ 📜Accordion.vue
>  ┃ ┃ ┃ ┃ ┣ 📜AccordionContent.vue
>  ┃ ┃ ┃ ┃ ┣ 📜AccordionItem.vue
>  ┃ ┃ ┃ ┃ ┣ 📜AccordionTrigger.vue
>  ┃ ┃ ┃ ┃ ┗ 📜index.js
>  ┃ ┗ 📂lib
>  ┃ ┃ ┗ 📜utils.js
> ```

Thank you for taking the time to test this!!

Yes, We may need to discuss this and how to solve it. 

When typescript is not enabled, we don't have `tsconfig`, which prevents the alias from being parsed correctly.

sadeghbarati:
❤️ 

What about `jsconfig.json`?

Dunqing:
I think so not good, most javascript projects don’t have `jsconfig.json`. Alias depends on the bundler tool.

zernonia:
Nicely done @Dunqing !

> When typescript is not enabled, we don't have `tsconfig`, which prevents the alias from being parsed correctly.

What about let's add another field during the `Setup` to let user enter their root path. For **Vue** is should be `/src`, and **Nuxt** should be `./` if there are not using monorepo

Dunqing:
> What about let's add another field during the `Setup` to let user enter their root path. For **Vue** is should be `/src`, and **Nuxt** should be `./` if there are not using monorepo

Sounds good to me, I will try it.

Dunqing:
It seems like we can't support aliases when typescript is disabled. 

Here is my `components.json`
```ts
{
  "root": "src",
  "style": "default",
  "typescript": false,
  "tailwind": {
    "config": "tailwind.config.js",
    "css": "src/assets/index.css",
    "baseColor": "zinc",
    "cssVariables": true
  },
  "framework": "vue",
  "aliases": {
    "components": "./components",
    "utils": "./lib/utils"
  }
}
```

But the component still contains the aliases import


image

Suppose we want to support aliases in our javascript. It's not enough to have a single root parameter, we need the user to provide the actual path to the components and utils.

Also, I found that `shadcn-ui` still needs `tsconfig.json` when `typescript` is disabled. Can we also do so that `tsconfig.json` is required? Whether you use typescript or not

zernonia:
@Dunqing  the reason I suggested adding a `src` folder is so that Shadcn components could be imported using relative path, instead of `alias`.

eg:
```
import { cn } from "../lib/utils"
```

zernonia:
@Dunqing  after looking at different approach, let's go with `jsconfig`. similar to `ui.shadcn`. I will improve the "installation" page for `shadcn-vue` to mentioned this

image

Dunqing:
> @Dunqing after looking at different approach, let's go with `jsconfig`. similar to `ui.shadcn`. I will improve the "installation" page for `shadcn-vue` to mentioned this

Does this mean we don't need the root option?

zernonia:
> > @Dunqing after looking at different approach, let's go with `jsconfig`. similar to `ui.shadcn`. I will improve the "installation" page for `shadcn-vue` to mentioned this
> 
> Does this mean we don't need the root option?

yup yup.

Dunqing:
I think we can move on to review.

zernonia (approved):
lgtm!

PR #47, opened 2023-09-13
chore: add todos in README.md

kedniko:


zernonia:
Sorry @kedniko  we dont need this table here, as it is solely dpends on the API and accessbility of radix-vue. Closing this. Thanks

PR #48, opened 2023-09-13
feat: add `taze`

sadeghbarati:
Added two script in root `package.json`

```
1. "taze": "taze major -frI --ignore-paths ./packages/cli/test/** --exclude typescript,/@iconify/",
2. "taze:minor": "taze minor -fwri --ignore-paths ./packages/cli/test/** --exclude /@iconify/"
```

1. with `taze` (major) script interactive mode `(I)` is on and it will prompt about write and installing packages at the end

2. with `taze:minor` it will bump and install packages automatically


zernonia:
@sadeghbarati  does Taze has a bot in GitHub that will automatically create PR when there's deps needing update?

sadeghbarati:
I don't think so
This is a temporary solution until you add some powerful GitHub actions like Renovate or Dependabot

zernonia:
@sadeghbarati  can you resolve those conflict? Then we can get this in! 😁

zernonia (approved):

PR #59, opened 2023-09-18
feat: add command palette

ahmedmayara:
I just added the `Command Palette` to allow the users to navigate through the docs more easily. Hope it helps 😁


apps/www/src/lib/registry/default/ui/command/CommandGroup.vue
```diff
@@ -11,7 +11,7 @@ const props = defineProps<ComboboxGroupProps & {
 <template>
   <ComboboxGroup
     v-bind="props"
-    :class="cn('overflow-hidden p-1 text-foreground', $attrs.class ?? '')"
```
zernonia, line 14: thanks for the PR @ahmedmayara . Howewer can we try not to edit styling in registry? We wanna keep the same styling as ui.shadcn's components.

apps/www/src/lib/registry/default/ui/dialog/DialogContent.vue
```diff
@@ -33,9 +33,9 @@ const emitsAsProps = useEmitAsProps(emits)
       <slot />
 
       <DialogClose
-        class="absolute top-4 right-4 p-0.5 transition-colors rounded-md hover:bg-secondary"
+        class="absolute top-3 right-3 p-0.5 transition-colors rounded-md hover:bg-secondary"
       >
-        <X class="w-4 h-4" />
+        <X class="w-4 h-4 text-muted-foreground" />
```
zernonia, line 38: same goes to here

apps/www/src/lib/registry/default/ui/command/CommandGroup.vue
ahmedmayara: I removed the overflow-hidden class because the command palette doesn't work properly, the scrollbar doesn't show, which means the users can't navigate through the command component.

PR #65, opened 2023-09-20
feat(cli): support adding all components

Dunqing:
Follow up https://github.com/shadcn-ui/ui/pull/1033


zernonia (approved):
Thanks @Dunqing !

PR #66, opened 2023-09-20
feat(cli): add overwrite confirmation for existing components

Dunqing:
Follow up https://github.com/shadcn-ui/ui/pull/973


zernonia:
Nice! Thanks for catching these feature from ui.shadcn!!! 😁

Issue #39, opened 2023-09-09
[Bug]: [vite-node] [plugin:vite:vue] [VITE_ERROR] /components/ui/avatar/AvatarImage.vue


selemondev:
### Environment

```bash
Developement/Production OS: MacOs
Node version: v18.16.1
Package manager: pnpm@8.6.7
Radix Vue version: 0.1.32
Vue version: 3.0.0
Nuxt version: 3.7.1
Nuxt mode: universal
Nuxt target: server
CSS framework: tailwindcss@3.3.3
Browser: Chrome 90.0.4430.212
```

### Link to minimal reproduction

https://stackblitz.com/edit/nuxt-starter-wbuhpq?file=app.vue

### Steps to reproduce

1. Installing a new component using the `shadcn-vue` CLI: npx shadcn-vue@latest add button

2. After the button has been successfully installed, it maps to the wrong path: `@/lib/registry/default/ui/avatar` and then I correct it to: `@/components/ui/avatar`

3. I go ahead and use it in my `app.vue` file as shown below : 

```js
<script setup lang="ts">
import { Avatar, AvatarFallback, AvatarImage } from '@/components/ui/avatar'
</script>

<template>
  <Avatar>
    <AvatarImage src="https://github.com/shadcn.png" alt="@shadcn" />
    <AvatarFallback>CN</AvatarFallback>
  </Avatar>
</template>
```

4. When I start my dev server by running `pnpm dev` it throws this error:  

[08:07:23]  WARN  [nuxt] Two component files resolving to the same name UiAvatar:

 - /home/projects/nuxt-starter-4b4eme/components/ui/avatar/index.ts
 - /home/projects/nuxt-starter-4b4eme/components/ui/avatar/Avatar.vue

7. When I navigate to `localhost:3000`, I get the following error: 

500
[vite-node] [plugin:vite:vue] [VITE_ERROR] /components/ui/avatar/AvatarImage.vue

<script setup lang="ts">
import { AvatarImage, type AvatarImageProps } from 'radix-vue'

const props = defineProps<AvatarImageProps>()
</script>

<template>
  <AvatarImage v-bind="props" class="h-full w-full object-cover" />
</template>

at /components/ui/avatar/AvatarImage.vue

### Describe the bug

Whenever I add a new component using the `shadcn-vue` CLI and start up my server, I get the following errors: 

[08:07:23]  WARN  [nuxt] Two component files resolving to the same name UiAvatar:

[...]

PR #71, opened 2023-09-22
docs: update Nuxt 3 installation guide

selemondev:
This pull request is intended to update the Nuxt 3 installation guide by adding `TypeScript` as a dependency to resolve issue [nuxt/nuxt/#20936](https://github.com/nuxt/nuxt/issues/20936) as discussed in issue #39

### DIFF
--- a/apps/www/src/content/docs/installation/nuxt.md
+++ b/apps/www/src/content/docs/installation/nuxt.md
@@ -13,6 +13,12 @@ Start by creating a new Nuxt project using `create-next-app`:
 npx nuxi@latest init my-app
 ```
 
+### Install TypeScript
+
+
+```bash
+npm install -D typescript
+```
 ### Install TailwindCSS module
 
 ```bash
@@ -30,7 +36,7 @@ export default defineNuxtConfig({
       // this is required else Nuxt will autoImport `.ts` file
       extensions: ['.vue'],
       // prefix for your components, eg: UiButton
-      prefix: 'Ui'
+      prefix: 'UI'
     },
   ],
 })
@@ -116,7 +122,7 @@ The command above will add the `Button` component to your project. Nuxt autoImpo
 ```vue {3}
 <template>
   <div>
-    <UiButton>Click me</UiButton>
+    <UIButton>Click me</UIButton>
   </div>
 </template>
 ```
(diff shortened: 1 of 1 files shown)


apps/www/src/content/docs/installation/nuxt.md
```diff
@@ -30,7 +36,7 @@ export default defineNuxtConfig({
       // this is required else Nuxt will autoImport `.ts` file
       extensions: ['.vue'],
       // prefix for your components, eg: UiButton
-      prefix: 'Ui'
+      prefix: 'UI'
```
zernonia: can we keep `Ui` as prefix? 😁

apps/www/src/content/docs/installation/nuxt.md
```diff
@@ -13,6 +13,12 @@ Start by creating a new Nuxt project using `create-next-app`:
 npx nuxi@latest init my-app
 ```
 
+### Install TypeScript
+
+
+```bash
+npm install -D typescript
+```
```
zernonia: Can we add some more info why this is needed? It only happened in certain case. I've tested Nuxt locally and it works without `typescript` deps. 

We can do this:

> if you encountered this issue `ERROR  Cannot read properties of undefined (reading 'sys') (x4)`, then installed Typescript as dependency. [Linked issue](https://github.com/nuxt/nuxt/issues/20936)

zernonia (changes requested):

apps/www/src/content/docs/installation/nuxt.md
selemondev: I was actually thinking of this but I thought that it would make the installation section a little bit bloated.

selemondev:
Done.

apps/www/src/content/docs/installation/nuxt.md
selemondev: Sure

apps/www/src/content/docs/installation/nuxt.md
zernonia: I think clearer instruction for installations is important, and it's just a few line, won't add too much bloat 😂

zernonia:
Thanks for the PR @selemondev !!!!

PR #83, opened 2023-09-24
Feature/avoid collisions

maelp:


zernonia:
Thanks for this PR @maelp . However in  `shadcn-vue^0.3.0` we've fixed the issue whereby the default `avoidCollisions` and some other props not populating correctly. Thus I'm closing this. Thanks 😁

maelp:
Perfect!

BTW what’s the best way to keep tabs on upstream since shadcn is made to be copy pasted? How do you do it yourself? Manual diffs from time to time?

PR #93, opened 2023-10-01
fix(components): add emits to the tabs component

leogallot:
I noticed that the `Tabs` component doesn't forward the emits from the `TabsRoot` component of `radix-vue`.


zernonia (approved):
Nice catch @leogallot !!! Thanks for the PR ya! Appreciate it

PR #94, opened 2023-10-01
Update select.md with correct docs link to select component

beejaz:
The link to vue-shadcn source & radix-vue docs was pointing to Popover component instead of Select component.

Not sure why it sees line 48 as a diff though?


zernonia (approved):
nice catch @beejaz ! thanks for the PR!

PR #108, opened 2023-10-09
docs: add `@formkit/auto-animate` example in form.md

sadeghbarati:


zernonia:
Nice showcase for the integration with `auto-animate`!!

PR #109, opened 2023-10-09
feat: add `pagination`

sadeghbarati:
~~no wrapper component needed for radix-vue pagination components~~

### DIFF
--- a/apps/www/src/lib/registry/new-york/example/PaginationDemo.vue
+++ b/apps/www/src/lib/registry/new-york/example/PaginationDemo.vue
@@ -1,0 +1,37 @@
+<script setup lang="ts">
+import {
+  Pagination,
+  PaginationEllipsis,
+  PaginationFirst,
+  PaginationLast,
+  PaginationList,
+  PaginationListItem,
+  PaginationNext,
+  PaginationPrev,
+} from '@/lib/registry/new-york/ui/pagination'
+
+import {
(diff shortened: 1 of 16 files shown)


apps/www/src/lib/registry/new-york/example/PaginationDemo.vue
```diff
@@ -0,0 +1,58 @@
...
+} from '@/lib/registry/new-york/ui/button'
+</script>
+
+<template>
+  <Pagination v-slot="{ page }" :total="100" :sibling-count="1" show-edges :default-page="2">
+    <PaginationList v-slot="{ items }" class="flex items-center gap-1">
+      <PaginationFirst as-child>
+        <Button class="w-9 h-9 p-0" variant="outline">
+          <DoubleArrowLeftIcon />
+        </Button>
```
zernonia: I think we can wrap these inside a wrapper component. As I observed, many of `shadcn-ui`'s icon were wrapped inside wrapper component.

I believe it would improve DX as dev just need to import the relevant component without needing to add the icon again.

WDYT?

sadeghbarati: Correct 💯 

I'll do it

Also thanks for the invitation ❤️

zernonia (approved):
Well done!! LGTM!

PR #116, opened 2023-10-13
fix: selecttrigger disabled style

olemarius:
changed tw prefix disabled: to data-[disabled]: to properly target selecttrigger

### DIFF
--- a/apps/www/src/lib/registry/default/ui/select/SelectTrigger.vue
+++ b/apps/www/src/lib/registry/default/ui/select/SelectTrigger.vue
@@ -17,7 +17,7 @@ const props = withDefaults(
     v-bind="props"
     :class="[
       cn(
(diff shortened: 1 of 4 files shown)


sadeghbarati:
Hi 

Thank you for your PR

I think the native `disabled` attribute is better than `data-disabled`, cause `focus-ring` still appears after clicking `SelectTrigger`

It should be resolved in [radix-vue](https://github.com/radix-vue/radix-vue/blob/main/packages/radix-vue/src/Select/SelectTrigger.vue#L62) (another disabled attribute beside data-disabled)



https://github.com/radix-vue/shadcn-vue/assets/17789047/8588ebd5-112b-4508-a039-624b40f81b39

sadeghbarati:
Radix V1 PR will resolve this issue

https://github.com/radix-vue/radix-vue/pull/436/files#diff-fec76c199a81b35187f26603a61169d86a8186fd74e8d749818e9a6f19e8f732

PR #117, opened 2023-10-13
fix: add v-bind="$attrs" to DatePicker in Calendar

olemarius:
Adding v-bind="$attrs" to DatePicker allows fine-grained control over the underlying v-calendar component's API.


sadeghbarati:
Attribute Inheritance is **true** by default

With your change attributes will apply to these two lines 👇  instead of only `<DatePicker>` component

- https://github.com/olemarius/shadcn-vue/blob/calendar-attrs/apps/www/src/lib/registry/default/ui/calendar/Calendar.vue#L55
- https://github.com/olemarius/shadcn-vue/blob/calendar-attrs/apps/www/src/lib/registry/default/ui/calendar/Calendar.vue#L65

sadeghbarati:
IMO, let's wait for [__Infer attrs (Vue 3.4)__](https://github.com/rudyxu1102/rfcs/blob/feat/infer_attrs/active-rfcs/0000-infer-attrs.md) cause `v-calendar` does not provide any types to set it easily in the defineProps section

olemarius:
@sadeghbarati thanks for pointing that out! Also agree it's better to wait for Infer attrs, wasn't aware of that rfc. 

So for now, the workaround is to modify your own local copy of the Calendar.vue component with

```js
defineOptions({
  inheritAttrs: false
})
```

and 

```html
<DatePicker 
  v-bind="$attrs" 
  ...  
/>
```

zernonia (approved):
lgtm!

PR #118, opened 2023-10-14
docs(app): add Astro guide

### LINKED ISSUE #90: [Question] Is all this compatible with Astro + Vue?
Hey i'm using astro.build right now and i was wondering if this is compatible, do i just have to install like it's nuxt? ty :)

selemondev:
This pull request is intended to add a guide on how to install, configure and use Shadcn-Vue with Astro.

Closes #90


sadeghbarati:
//cc @Dunqing 🙏 

- ```pnpm dlx astro add tailwind```
- ```pnpm dlx shadcn-vue@latest init```

image

---

Prompt texts are also different

image
image

selemondev:
@sadeghbarati Does this pull request solve the issue or does it need more commits?

sadeghbarati:
Your PR is good but `shadcn-vue` CLI need some change when the framework is `Astro`

let's merge when everything is ok 🙏

selemondev:
> Your PR is good but `shadcn-vue` CLI need some change when the framework is `Astro`
> 
> let's merge when everything is ok 🙏

What change, if I may ask?

sadeghbarati:
I explained with the images above 👆 

- `shadcn-vue` CLI created **tailwind.config.js**
- `astro` CLI created **tailwind.config.mjs**

selemondev:
> I explained with the images above 👆
> 
> * `shadcn-vue` CLI created **tailwind.config.js**
> * `astro` CLI created **tailwind.config.mjs**

Will this be opened as an issue?

selemondev:
Is there any valid reason that we can't use the tailwind config created by the `astro-tailwind` plugin (tailwind.config.mjs)? I think the CLI doesn't have an issue.

Dunqing:
I think both `tailwind.config.js` and `tailwind.config.mjs` should be supported in shadcn-vue. I will support it soon.

zernonia:
Nice suggestion guys! As different modules add tailwind differently, it's good that we cover most of the use case for `tailwind.config.{js,cjs,mjs}`. I believe https://github.com/radix-vue/shadcn-vue/pull/120 would work nicely!

zernonia:
Thanks guys! I've merge #120  and released `0.6.0`. @selemondev  could you test it on `astro` and check if everything is working as expected? Then we can merge this 😁

selemondev:
> Thanks guys! I've merge #120 and released `0.6.0`. @selemondev could you test it on `astro` and check if everything is working as expected? Then we can merge this 😁

I have tested it out but after the installation and adding of the Shadcn components, I had to install TypeScript as a dev dependency to get it working and resolve the error `Cannot find module 'typescript'`. It is the same case as Nuxt 3. Can I go ahead and add that in the documentation as well (Installing TypeScript as a dev dependency)?

zernonia:
Yes @selemondev . Let's add that step!

selemondev:
> Yes @selemondev . Let's add that step!

Done 🥳.

PR #120, opened 2023-10-15
feat(cli): support `tailwind.config.mjs` file

Dunqing:
When the extension of the tailwind configuration file is mjs, transform cjs syntax to esm syntax.

If I understand it right, we just need to handle tailwind.config.mjs file correctly in `astro` framework? 
cc @sadeghbarati @selemondev

### DIFF
--- a/packages/cli/src/commands/init.ts
+++ b/packages/cli/src/commands/init.ts
@@ -28,6 +28,7 @@ import {
   rawConfigSchema,
   resolveConfigPaths,
 } from '../utils/get-config'
+import { transformCJSToESM } from '../utils/transformers/transform-cjs-to-esm'
 
 const PROJECT_DEPENDENCIES = {
   base: [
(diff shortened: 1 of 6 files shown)


sadeghbarati:
Yes, Right `@astrojs/tailwind>=@5.0.2` integration will create `tailwind.config.mjs`

https://github.com/withastro/astro/blob/main/packages/integrations/tailwind/CHANGELOG.md#502

sadeghbarati:
Can I push some changes? @Dunqing

Dunqing:
> Can I push some changes? @Dunqing 

Sure！

zernonia:
Thanks for the PR @Dunqing  @sadeghbarati ! However seems like the Test in the pipeline has failed. Could you check on that? 😁

Dunqing:
> Thanks for the PR @Dunqing @sadeghbarati ! However seems like the Test in the pipeline has failed. Could you check on that? 😁

Done

PR #125, opened 2023-10-20
fix: customizer code text color

XavierVilaRubio:
Added text-white on code lines


zernonia:
@XavierVilaRubio  why is this needed ya? Does it render differently for you in shadcn-vue.com?

image

sadeghbarati:
Other shadcn sites are using multiple `text-white` classes for line code instead of one `text-white` on parent element (code tag)
for consistency let's keep it the same as other shadcn sources 😄 

image

sadeghbarati (approved):

PR #135, opened 2023-10-22
dynamic text direction

farajshuaib:
change text direction from statically left to be justifying parent element, which will be right if the direction is rtl and left if the direction is ltr

### DIFF
--- a/apps/www/.vitepress/theme/layout/MainLayout.vue
+++ b/apps/www/.vitepress/theme/layout/MainLayout.vue
@@ -171,7 +171,7 @@ watch(() => $route.path, (n) => {
 
     <footer class="py-6 md:px-8 md:py-0">
       <div class="container flex flex-col items-center justify-between gap-4 md:h-24 md:flex-row">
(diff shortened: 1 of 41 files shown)


sadeghbarati:
@farajshuaib 

Thanks for PR but I don't think only `text-justify` could help respect the direction of document

CSS Logical Properties and gaps are the way, for reference, you can search **`RTL support`** in main shadcn-ui GitHub

farajshuaib:
> @farajshuaib
> 
> Thanks for PR but I don't think only `text-justify` could help respect the direction of document
> 
> CSS Logical Properties and gaps are the way

nono I'm not talking about the direction of the document it's just the text alignment itself, it's a problem that happens with me when I start a project that depends on Chadcn UI components and needs RTL direction all text alignment is justified with the direction of the document except the component that I've used from shadcn UI  are statically aligned left because there's a class scoped it to be aligned to the left which is 'text-left'.
it's easy to customize it by the way but it will be good if it's already aligned left or right by the parent  element

PR #138, opened 2023-10-23
chore: upgrade `www` deps

sadeghbarati:


sadeghbarati:
Will request to review next time for major deps

Issue #146, opened 2023-10-31
[Feature]: Ability to customize globally the @/components & @/utils paths in documentation


TechAkayy:
### Describe the feature

Hi team, 

As `@/components` is the default path set when scaffolding, having the same path displayed in the component code could help us easily copy paste. 

To take it to the next level, it would be great to have an option to set these two paths globally across the documentation, so that users can set their paths at their end, and copy the code without having to amend the paths at our ends. Thank you! 

![image](https://github.com/radix-vue/shadcn-vue/assets/73772701/e753e18f-a02a-46d2-a5d6-471a4ee2cb30)

### Additional information

- [ ] I intend to submit a PR for this feature.
- [ ] I have already implemented and/or tested this feature.

sadeghbarati:
@TechAkayy 

Fixed `@/components/ui/***` in my fork

Managed to add dynamic alias, but it only works in `<ComponentPreview>`
It is hard to change rendered markdown content, for example, go to `alert-dialog.md`



```
#### markdown:

```vue
<script setup lang="ts">
import {
  AlertDialog,
  AlertDialogAction,
  AlertDialogCancel,
  AlertDialogContent,
  AlertDialogDescription,
  AlertDialogFooter,
  AlertDialogHeader,
  AlertDialogTitle,
  AlertDialogTrigger,
} from '@/components/ui/alert-dialog'
</script>

<template>
  <AlertDialog>
    <AlertDialogTrigger>Open</AlertDialogTrigger>
    <AlertDialogContent>
      <AlertDialogHeader>
        <AlertDialogTitle>Are you absolutely sure?</AlertDialogTitle>
        <AlertDialogDescription>
          This action cannot be undone. This will permanently delete your account
          and remove your data from our servers.
        </AlertDialogDescription>
      </AlertDialogHeader>
      <AlertDialogFooter>
        <AlertDialogCancel>Cancel</AlertDialogCancel>
        <AlertDialogAction>Continue</AlertDialogAction>
      </AlertDialogFooter>
    </AlertDialogContent>
  </AlertDialog>
</template>
```
```

#### rendered:


[...]

PR #147, opened 2023-11-02
feat: add `pinia` and `pinia-shared-state` for sync shared state between tabs

sadeghbarati:
States like

- Color
- Radius
- Style

### DIFF
--- a/apps/www/.vitepress/theme/components/ComponentLoader.vue
+++ b/apps/www/.vitepress/theme/components/ComponentLoader.vue
@@ -1,4 +1,5 @@
 <script setup lang="ts">
+import { storeToRefs } from 'pinia'
 import { defineAsyncComponent } from 'vue'
 import Spinner from './Spinner.vue'
 import { useConfigStore } from '@/stores/config'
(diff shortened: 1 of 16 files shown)


zernonia:
curious what was the reason to use pinia 🤔

sadeghbarati:
You were using `@vueuse/core` `useSession` for states
Every `Browser Tab` has its own session and states are not shared across tabs

sadeghbarati:
Kinda related https://github.com/radix-vue/shadcn-vue/issues/146

I was working on adding custom alias (`@/components/`) in https://www.shadcn-vue.com/themes.html (Customize Button) but could only handle changing aliases in `<ComponentPreview>` component, not in rendered markdowns 

then I tried sharing other states

zernonia:
we can just swap it with `useLocalStorage` lol. I dont think we need full-fledged state management for this tiny config store


https://github.com/radix-vue/shadcn-vue/assets/59365435/7c243afd-a725-4e31-9704-35db92f89aac

sadeghbarati:
You right 😂

For Aliases I thought better to turn it back to default alias when user close all shadcn-vue tabs 

btw I removed dynamic alias implementation before pusing changes 😭😁

feel free to close 🫡

PR #155, opened 2023-11-09
fix: Fixed announcement typo

lovlyx:


zernonia (approved):
Thanks for catching that!

Issue #174, opened 2023-11-17
[Bug]: new-york's CommandList missing <div role="presentation">


anonimusprogramus:
### Environment

```bash
Developement/Production OS: Linux Mint 20.1 Cinnamon
Node version: v18.16.0
Package manager: pnpm@8.9.2
Radix Vue version: 1.1.1
Shadcn Vue version: 0.8.2
Vue version: 3.3.7
Nuxt version: 3.8.0
Nuxt mode: universal
Nuxt target: client server
CSS framework: tailwindcss@3.3.3
Client OS: Linux Mint 20.1 Cinnamon
Browser: Brave 1.60
```

### Link to minimal reproduction

https://github.com/radix-vue/shadcn-vue/blob/dev/apps/www/src/lib/registry/new-york/ui/command/CommandList.vue

### Steps to reproduce

- Clone shadcn-vue.com but with style set to "**new-york**"
- Open Command-K or "**Search for anything..." at homepage**
- Dialog will show
- Command list viewport is not scrollable

![2023-11-17 16-48-34](https://github.com/radix-vue/shadcn-vue/assets/7464749/014a10c9-311f-4209-afd9-a8fb12557f94)

### Describe the bug

A `<div role="presentation">` is missing from "**new-york**" but exists in "**default**"

A PR will be submitted shotly

### Expected behavior

![2023-11-17 16-48-59](https://github.com/radix-vue/shadcn-vue/assets/7464749/da921b1b-2179-4b76-b97a-3f4ec160d428)

PR #175, opened 2023-11-17
Update CommandList.vue

anonimusprogramus:
A `<div role="presentation">` is missing from "new-york" but exists in "default"

Fixing #174


sadeghbarati (approved):
🙏

PR #177, opened 2023-11-17
refactor: move InjectionKey export from FormItem.vue to .ts file to avoid warnings

### LINKED ISSUE #176: [Feature]: Move InjectionKey export from FormItem.vue to .ts file to avoid warnings
### Describe the feature

Today useFormField.ts imports FORM_ITEM_INJECTION_KEY from FormItem.vue. 

In a monorepo with a /package/my-lib with shadcn-vue set up, this gives the following error when used in an app in the same monorepo:

image

I've added the `'import/first': 'off' ` rule both in the library's eslint config and the app importing the library's eslint config with no luck. My local fix for this is to move the InjectionKey out to a separate .ts file and import it from there.

/ui/form/InjectionKey.ts:

```js
import { type InjectionKey } from 'vue'

export const FORM_ITEM_INJECTION_KEY
  = Symbol() as InjectionKey<string>
```

I'm not sure why I can't import from vue files, but I think using a .ts file might be a better approach anyways.

Let me know if I should create a PR.  

### Additional information

olemarius:
Fixes #176

### DIFF
--- a/apps/www/src/lib/registry/default/ui/form/FormItem.vue
+++ b/apps/www/src/lib/registry/default/ui/form/FormItem.vue
@@ -1,10 +1,3 @@
-<script lang="ts">
-import { type InjectionKey } from 'vue'
-
-export const FORM_ITEM_INJECTION_KEY
-  = Symbol() as InjectionKey<string>
-</script>
-
 <script lang="ts" setup>
 import { provide, useAttrs } from 'vue'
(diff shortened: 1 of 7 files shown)


olemarius:
Failing to import from *.vue files happened because I was using tsc instead of vue-tsc. See: 

https://github.com/radix-vue/shadcn-vue/issues/176#issuecomment-1817746656

PR #179, opened 2023-11-17
refactor: use pathe instead of node:path

### LINKED ISSUE #178: use pathe instead of node:path
### Describe the feature

build registry fails on windows because the paths are \\ instead of /, so split on "/example" returns undefined. 

image

I'll create a PR where I replace node:path with pathe, ref https://github.com/unjs/pathe

### Additional information

olemarius:
Closes #178


zernonia:
Thanks for the PR @olemarius  😁

PR #181, opened 2023-11-17
Output consistency across platforms

### LINKED ISSUE #180: [Feature]: output consistency across platforms
### Describe the feature

Running build:registry on windows generates \r\n newlines, while unix based systems generates \n, and readdir does not have consistent sorting. I've added .sort(), but I'm unable to verify that this will put index on top on unix based system. 

The behavior today is putting index.ts in the bottom of the file list regardless of other filenames in list. 

Great if someone could confirm that the .sort() is applied on unix. 

### Additional information

olemarius:
Fixes https://github.com/radix-vue/shadcn-vue/issues/180

### DIFF
--- a/apps/www/package.json
+++ b/apps/www/package.json
@@ -11,7 +11,7 @@
     "preview": "vitepress preview",
     "typecheck": "vue-tsc --noEmit",
     "typecheck:registry": "vue-tsc --noEmit -p tsconfig.registry.json",
-    "build:registry": "npm run typecheck:registry && tsx ./scripts/build-registry.ts"
+    "build:registry": "pnpm typecheck:registry && tsx ./scripts/build-registry.ts"
   },
   "dependencies": {
     "@formkit/auto-animate": "^0.8.0",
--- a/apps/www/package.json
+++ b/apps/www/package.json
@@ -11,7 +11,7 @@
     "preview": "vitepress preview",
     "typecheck": "vue-tsc --noEmit",
     "typecheck:registry": "vue-tsc --noEmit -p tsconfig.registry.json",
-    "build:registry": "typecheck:registry && tsx ./scripts/build-registry.ts"
+    "build:registry": "npm run typecheck:registry && tsx ./scripts/build-registry.ts"
   },
   "dependencies": {
     "@formkit/auto-animate": "^0.8.0",
--- a/apps/www/scripts/build-registry.ts
+++ b/apps/www/scripts/build-registry.ts
@@ -66,6 +66,8 @@ fs.writeFileSync(path.join(process.cwd(), '__registry__/index.ts'), index)
 // ----------------------------------------------------------------------------
 // Build registry/styles/[style]/[name].json.
 // ----------------------------------------------------------------------------
+const newLine = '\n'
+
 for (const style of styles) {
   const targetPath = path.join(REGISTRY_PATH, 'styles', style.name)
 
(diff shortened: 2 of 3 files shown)


sadeghbarati:
Can you provide images of the before and after the change you did? Thanks

This would make it easy for other maintainers to review

olemarius:
newline inconsistency (fixed in this PR)
image

file order inconsistency (fixed in this PR, but needs to be confirmed by someone on unix based platform):
image

Also, quotes inconsistency (NOT fixed in this PR)
image

sadeghbarati:
That is eslint problem we should migrate to v2 of antfu configuration 

and ignore registry build files in eslint.config.js

olemarius:
@sadeghbarati ok, so windows users should just build registry and push even with windows newline and other inconsistencies ?

sadeghbarati:
I'm not familiar with node too much but I saw the pain of windows path in other projects 🫠

olemarius:
@sadeghbarati yea me neither but thanks to unjs/pathe it was super easy to swap, ref https://github.com/radix-vue/shadcn-vue/pull/179 . AFAIK unjs has a bunch of tools like this to make it easier to deal with different platforms.

zernonia:
Yeah the registry script was designed to be run on unix based operating system. I've checkout the branch and ran the registry, seems to be getting the same result as on `dev` branch. @olemarius  I believe this change would allow windows user to run the script, and not affecting unix user right? 😁

olemarius:
@zernonia from what I could see, the index.ts file was always in the bottom of the file list, even when the other files started with a letter _after_ the letter "i".  

I added .sort() that should sort it alphabetically both on windows and unix, but if you're getting the same results on dev I'm not sure if the sort is applied ? 

There's an issue posted on readdir Windows sorting inconsistency here https://github.com/nodejs/node/issues/3232

sadeghbarati:
@olemarius You are Windows user right? If you want to test on unix system just open you GitHub fork in StackBlitz playground

https://stackblitz.com/github/olemarius/shadcn-vue/tree/output-consistency-across-platforms

olemarius:
Did some adjustment and found that .sort() will put lowercase after uppercase, from chatgpt:

_The reason 'index.ts' is placed last in the sorted array is because JavaScript's sort() method uses Unicode code point order to sort strings by default. In Unicode, uppercase letters come before lowercase letters, and special characters and numbers come before letters. In this case, 'index.ts' starts with a lowercase letter ('i'), and the other strings start with uppercase letters ('T'). That's why 'index.ts' comes after the other strings when sorted in Unicode order._

index.ts is now _last_ both on Windows and on Unix (tested on StackBlitz), so that Windows users will update only files  that actually changed. 

Windows:
image

But www/__registry__/index.ts uses double quotes both on stackblitz and windows. Is it converted to to singleqotes via lint staged?

apps/www/package.json
```diff
@@ -11,7 +11,7 @@
     "preview": "vitepress preview",
     "typecheck": "vue-tsc --noEmit",
     "typecheck:registry": "vue-tsc --noEmit -p tsconfig.registry.json",
-    "build:registry": "typecheck:registry && tsx ./scripts/build-registry.ts"
+    "build:registry": "npm run typecheck:registry && tsx ./scripts/build-registry.ts"
```
sadeghbarati: I forgot to add `pnpm`, change `npm run` to `pnpm` please

```suggestion
    "build:registry": "pnpm typecheck:registry && tsx ./scripts/build-registry.ts"
```

sadeghbarati (approved):
Finally tested and checked with Source Control changes, it had the same output as Unix-based environments.

PR #182, opened 2023-11-19
add vue-tsc

olemarius:
https://github.com/radix-vue/shadcn-vue/issues/176#issuecomment-1817967478


sadeghbarati:
```ts
import type { DatePickerModel } from 'v-calendar/dist/types/src/use/datePicker'
```

This will work only in `"moduleResolution": "Node"` the new `Bundler` mode will throw errors

```
Cannot find module 'v-calendar/dist/types/src/use/datePicker' or its corresponding type declarations.
```

---

Can you extract the type from `v-calendar` and put the type inside the `Calendar.vue`?

---

Looks like the main `shadcn-ui` is going to add the new Calendar implementation with `@internationalized/date` we can inspire from them once it's ready

olemarius:
@sadeghbarati ✅ done

sadeghbarati (approved):

sadeghbarati:
86nvki

PR #183, opened 2023-11-20
fix: card padding

olemarius:
CardHeader and CardFooter should be optional when using Card, and paddings should be correct regardless. 

I've applied peer class to only set pt-0 if a peer with pt-6 is present. 

```html
    <div>
      <Card :class="cn('w-[380px]', $attrs.class ?? '')">
        <CardHeader>
          <CardTitle>Title</CardTitle>
        </CardHeader>
        <CardContent> Content </CardContent>
        <CardFooter>Footer</CardFooter>
      </Card>
    </div>
      <div>
      <Card :class="cn('w-[380px]', $attrs.class ?? '')">
          <CardHeader>
          <CardTitle>Header only</CardTitle>
        </CardHeader>
      </Card>
    </div>
    <div>
      <Card :class="cn('w-[380px]', $attrs.class ?? '')">
        <CardContent> Content only </CardContent>
      </Card>
    </div>
    <div>
      <Card :class="cn('w-[380px]', $attrs.class ?? '')">
        <CardFooter>Footer only</CardFooter>
      </Card>
    </div>
    <div>
      <Card :class="cn('w-[380px]', $attrs.class ?? '')">
        <CardContent> Content </CardContent>
        <CardFooter>Footer</CardFooter>
      </Card>
    </div>
      <div>
      <Card :class="cn('w-[380px]', $attrs.class ?? '')">
        <CardHeader>
          <CardTitle>Title</CardTitle>
        </CardHeader>
        <CardContent> Content </CardContent>
      </Card>
    </div>
```

image

### DIFF
--- a/apps/www/src/lib/registry/default/ui/card/CardContent.vue
+++ b/apps/www/src/lib/registry/default/ui/card/CardContent.vue
@@ -10,7 +10,7 @@ const props = defineProps({
 </script>
 
 <template>
-  <div :class="cn('p-6 pt-0', props.class)">
+  <div :class="cn('p-6 peer peer-[.p-6]:pt-0', props.class)">
     <slot />
   </div>
 </template>
--- a/apps/www/src/lib/registry/default/ui/card/CardFooter.vue
+++ b/apps/www/src/lib/registry/default/ui/card/CardFooter.vue
@@ -10,7 +10,7 @@ const props = defineProps({
 </script>
 
 <template>
-  <div :class="cn('p-6 pt-0', props.class)">
+  <div :class="cn('p-6 peer-[.p-6]:pt-0', props.class)">
     <slot />
   </div>
 </template>
--- a/apps/www/src/lib/registry/default/ui/card/CardHeader.vue
+++ b/apps/www/src/lib/registry/default/ui/card/CardHeader.vue
@@ -10,7 +10,7 @@ const props = defineProps({
 </script>
 
 <template>
-  <div :class="cn('flex flex-col space-y-1.5 p-6', props.class)">
+  <div :class="cn('flex flex-col space-y-1.5 p-6 peer', props.class)">
     <slot />
   </div>
 </template>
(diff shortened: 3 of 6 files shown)


zernonia:
using `peer` would introduce some element of surprise for user.
Let's take a look at an example:

User might want to change the padding for `CardContent` with `p-8` instead of original `p-6` for 1 of the instance. When this case happen, the styling effect  of `peer-[.p-6]` will be gone and thus render the user by surprise with new extra padding.

olemarius:
@zernonia ok, how about adding some some non-utility classnames that only work as a "anchor" for peer? like .card-section

sadeghbarati:
@olemarius 

Just take a look at the main `shadcn-ui` card code, if the card's code is not ok search card in shadcn-ui `PRs`

olemarius:
@sadeghbarati is the philosophy to 'mirror' the original shadcn/ui as close as possible, so that if there's an enhancement or fix for shadcn-ui it should be applied there first? I did a quick test and it's the same issue there. If no CardHeader is present, the CardContent has no padding-top, see:

image

My PR intend to allow a more flexible use of Card, so if you choose to use it without CardHeader, CardContent will get the same padding on all sides. I'll be happy to suggest this for the original shadcn/ui project if that's the way to go forward with enhancements.

sadeghbarati:
Maybe `:has` CSS could help us here

https://twitter.com/adamwathan/status/1733176745082122746?t=PVbklcNDqm0-eukPD6yuUQ&s=19

sadeghbarati:
https://github.com/radix-vue/radix-vue/issues/505

Adding something like this to the Card components could help us solve this issue.


#### `Card.vue`
```html
<div data-radix-vue-card></div>
```

#### `CardDescription.vue`
```html
<div data-radix-vue-card-description>...</div>
```

sadeghbarati:
Will fix this issue later with CSS `:has`

PR #189, opened 2023-11-26
docs: add capability to user to use custom prefix or custom alias path for components and utils

### LINKED ISSUE #146: [Feature]: Ability to customize globally the @/components & @/utils paths in documentation

sadeghbarati:
### Draft until we fix duplicated auto-import component warn in Nuxt

Partially resolves https://github.com/radix-vue/shadcn-vue/issues/146 with https://vitepress.dev/guide/using-vue#unescape-in-code-blocks

<details><summary>Customize Button in /themes.html</summary>
<p>

image

</p>
</details>

<details><summary>Result</summary>
<p>

#### custom components alias path and custom prefix
image

#### custom utils alias path
image

</p>
</details>

---

Only `Usage` sections will have this feature.

Components that don't have this feature:

- Combobox
- DataTable
- Pagination

> Why?

Because of their complex code example that uses `v-for` and `v-if`, and since we using `vue-vue` to enable Vue-style interpolation inside fences, that will cause error in parsing markdown 

To solve the above problem we can use a simple Anatomy example without `v-for` and `v-if`

---

> I'm using auto-import I don't need those imports can you hide it when I enable the auto-import checkbox inside Customize Button? (not implemented)

We can hide it with code like this in md files

<details><summary>Details</summary>
<p>

image

</p>
</details>

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -2,8 +2,8 @@
   "prettier.enable": false,
   "editor.formatOnSave": false,
   "editor.codeActionsOnSave": {
-    "source.fixAll.eslint": true,
-    "source.organizeImports": false
+    "source.fixAll.eslint": "explicit",
+    "source.organizeImports": "never"
   },
 
   "eslint.validate": [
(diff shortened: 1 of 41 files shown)


sadeghbarati:
Will try to create new PR this is too old

PR #194, opened 2023-11-29
fix: input duplicated class name, remove type="text"

sadeghbarati:
If the **type** attribute is not specified, the default type is "text"


zernonia (approved):
LGTM!

PR #198, opened 2023-12-01
fix: theme switcher current colors

joshlarsen:
On the theme switcher, there are a few inconsistencies in the current color indicator:

- The border color of the current color indicator doesn't match the selected color. This is the behavior of the main [shadcn theme switcher](https://ui.shadcn.com/themes)
- The background color is also slightly off.
- The tooltip background color is that of the current color rather than a dark background

This PR fixes all of these.

Before:
Screenshot 2023-12-01 at 4 03 38 PM
Screenshot 2023-12-01 at 4 02 53 PM
Screenshot 2023-12-01 at 4 35 29 PM
Screenshot 2023-12-01 at 4 34 44 PM

After
Screenshot 2023-12-01 at 4 21 07 PM
Screenshot 2023-12-01 at 4 21 15 PM
Screenshot 2023-12-01 at 4 21 01 PM
Screenshot 2023-12-01 at 4 20 51 PM
Screenshot 2023-12-01 at 4 33 01 PM

### DIFF
--- a/apps/www/.vitepress/theme/layout/ThemingLayout.vue
+++ b/apps/www/.vitepress/theme/layout/ThemingLayout.vue
@@ -98,7 +98,7 @@ watch(radius, (radius) => {
                       class="flex h-9 w-9 items-center justify-center rounded-full border-2 border-border text-xs"
                       :class="
                         color === theme
-                          ? 'border-foreground'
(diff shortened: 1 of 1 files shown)


sadeghbarati (approved):
Thanks

PR #203, opened 2023-12-06
fix(Button): missing `asChild`

### LINKED ISSUE #202: [Bug]: Issue with Button as-child prop when using link
### Environment

```bash
Developement/Production OS: Windows 10 19043.1110
Node version: 16.0.0
Package manager: pnpm@8.6.0
Radix Vue version: 1.0.0
Shadcn Vue version: 1.0.0
Vue version: 3.0.0
Nuxt version: 3.0.0
Nuxt mode: universal
Nuxt target: server
CSS framework: tailwindcss@3.3.3
Client OS: Windows 10 19043.1110
Browser: Chrome 90.0.4430.212
```

https://www.shadcn-vue.com/docs/components/button.html#as-child

### Steps to reproduce

Using the as child prop doesnt work as expected 

```
<script setup lang="ts">
import { Button } from '@/components/ui/button'
</script>

<template>
  <Button as-child>
    <NuxtLink to="/login">
      Login
    </NuxtLink>
  </Button>
</template>
```

Doing it this way doesnt make the whole button clickable

### Describe the bug

When using as-child prop in Button and putting a Nuxt Link in the center the Link doesnt take up the width and height of the button by default i noticed this after testing on mobile

Heres a button i have in my navbar
image
when i inspect the css this is what it looks like
image
and here is the code
image

as you can see when passing the nuxt link as a child it doesnt take up the space of the entire button and is only clickable in the middle.

 one way around this is to style the link further and getting rid of the padding for the button because it adds padding on left and right
 
 here is another button in my app
image

 
 here is the css inspect as you see here its now not only in the middle and taking up the entire space
image

here is the code
image

the as child example on the shadcn vue site also has the same issue:
image

### Expected behavior

The Link inside the button should automatically wrap the entire link so that when you click anywhere in the button it should act as a link

_No response_

zernonia:


e-tho:
@zernonia,

Is this PR supposed to be present in shadcn-nuxt 0.8.6? `as-child` still has no effect in my case with: 
```
<Button variant="link" as-child>
  <NuxtLink to="/">
    Test
  </NuxtLink>
</Button>
```

zernonia:
You need to install `radix-vue`. You can manually install the pacakge, or run `npx shadcn-vue@latest init` again

sadeghbarati:
Can you share your project? @e-tho

e-tho:
@zernonia,

> You need to install `radix-vue`. You can manually install the pacakge, or run `npx shadcn-vue@latest init` again

With `radix-vue` manually installed all my instances of the `<Button>` component as well as those of the `<AccordionTrigger>` and `<AccordionContent>` components produce hydration errors: 
```
[Vue warn]: Hydration attribute mismatch on ...
- rendered on server: id="radix-75"
- expected on client: id="radix-1"
...
```
the result is the same when I initialize shadcn-vue again and `as-child` still has no effect

PR #211, opened 2023-12-16
Change docs link in sheet component

beejaz:
Component source linked in docs points to dialog component instead of sheet component.


sadeghbarati:
Thanks 🫡

PR #225, opened 2023-12-29
fix: allow .ts or .js extension in module package

Tiarhai:
### ❓ Type of change

### 📚 Description

In the `module` `package`, there is an issue related to the hardcoded reference to the `index.ts` file for each component.

The problem arises when opting not to use TypeScript while configuring the `components.json`. In such cases, upon adding a new `shadcn` component, the CLI generates an `index.js` file instead of a `index.ts` file.

Consequently, during project startup, an error occurs with the message `ERROR [unhandledRejection] File not found: C:/[...]/components/ui/avatar/index.ts` for each component (for clarity, the folder structure and the component name are on a per-installed component basis, and is not always `avatar`).

This pull request addresses this bug by introducing a fix that checks for either file (`index.ts` or `index.js`), preventing the aforementioned error during project initialization.

#### Before

Console On Startup - Before

#### After

Console On Startup - After

### DIFF
--- a/packages/module/src/module.ts
+++ b/packages/module/src/module.ts
@@ -1,4 +1,5 @@
 import { readdirSync } from 'node:fs'
+import { join } from 'node:path'
 import { addComponent, createResolver, defineNuxtModule } from '@nuxt/kit'
 import { Project } from 'ts-morph'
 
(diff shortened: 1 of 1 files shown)


packages/module/src/module.ts
```diff
@@ -1,4 +1,5 @@
 import { readdirSync } from 'node:fs'
+import { join } from 'node:path'
```
sadeghbarati: Thanks for PR, 

Can you change it from `node:path` to [`pathe`](https://github.com/unjs/pathe/blob/main/src/path.ts#L62) 
Since it's a dependency in `nuxt` and `nuxt/kit`, we can use it instead to prevent path problems

Tiarhai: @sadeghbarati done! 

And thank you for the amazing project.

sadeghbarati (approved):

packages/module/src/module.ts
sadeghbarati: It's Community / Team Work Project 🙏 🤩

sadeghbarati:
There is also a simple module which you can copy paste the code without adding another npm pkg

https://stackblitz.com/edit/github-xkryhq-egbi1f?file=modules%2Fshadcn-vue.ts,nuxt.config.ts

PR #229, opened 2023-12-29
docs: make `<DataTable />` reactive

jankremlacek:
The actual `<DataTable />` documentation example is not reactive. The PR fixes it.

The `<DataTable />` is an example component not included in the build, just in the documentation. The consumer of the `<DataTable />` component passes the `data` and the `columns` properties which are used in the `useVueTable()` composable from [TanStack Query](https://tanstack.com/query/latest/docs/vue/overview).

They would not work reactively as they are passed from the props. Getters fix it. Basically:

```diff
const table = useVueTable({
-    data: props.data,
+    get data() { return props.data },
    ...
})
```

### DIFF
--- a/apps/www/src/content/docs/components/data-table.md
+++ b/apps/www/src/content/docs/components/data-table.md
@@ -5,7 +5,7 @@ primitive: https://tanstack.com/table/v8/docs/guide/introduction
 ---
 
 
-<ComponentPreview name="DataTableDemo"  /> 
+<ComponentPreview name="DataTableDemo"  />
 
 ## Introduction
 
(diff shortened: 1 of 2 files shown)


sadeghbarati:
Thanks for the PR. Can you share an example? 
Looks like they have the same issue with the getters

https://github.com/TanStack/table/issues/4708

jankremlacek:
Sure, @sadeghbarati, here is the example: https://codesandbox.io/p/devbox/tanstack-usevuetable-using-getter-jhcr84

The example shows both ways to bind `props.data`, direct binding doesn't work, using the getter works.

> Note: The example does not use any of the `shadcn-vue` components, as the issue is with passing props to the `useVueTable()`.

sadeghbarati (approved):

PR #231, opened 2023-12-30
docs: change cards radius to `rounded-lg`

itsTPM:
currently, when changing the radius in [themes page](https://www.shadcn-vue.com/themes.html), the radius only changes on the first two cards because the `Card.vue` component has a rounded-xl class, which is wrong cuz it doesn't use `--radius` in the css
Before
cards with rounded-xl
After
cards with rounded-lg

p.s. `LinkedCard.vue` has also changed to use rounded-lg, so now if the user changes the radius, it will be applied to that element as well:
LinkedCard.vue


sadeghbarati:
Thanks 🫡

PR #235, opened 2024-01-02
fix(Module): HMR not working when edit component load too slow

### LINKED ISSUE #214: [Bug](Nuxt modules): HMR not working when edit component, load too slow
### Environment

```bash
Shadcn Vue version: latest
Nuxt version: 3.8.2
```

run locally

### Steps to reproduce

1. Fork https://github.com/zernonia/memory-lanes,
2. pnpm i
3. pnpm dev

### Describe the bug

image

Slow modules, took 7s to load many components.

### Expected behavior

HMR should work when edit component, faster module

_No response_

### LINKED ISSUE #222: [Bug]: CVA styling changes don't get applied without a server restart
### Environment

```bash
Development/Production OS: Mac Ventura 13.6
Node version: v21.0.0
Package manager: pnpm@8.6.0
Radix Vue version: 1.0.0
Shadcn Vue version: (shadcn-nuxt) 0.1.0
Vue version: 3.3.13
Nuxt version: 3.9.0
Nuxt mode: universal
Nuxt target: server
CSS framework: tailwindcss@6.10.1
```

local server required

### Steps to reproduce

-Start a new Nuxt project
-Install all shadcn requirements
-install a button component using CLI
-start local live server
-change the button/index.ts -> CVA styles (any)
-watch the browser and see that nothing has changed after hard browser refresh
-restart server
-confirm that changes are now live in the browser

### Describe the bug

I am unable to update CVA styling without having to also restart my local server for the changes to take effect.
Other changes to templates, etc, do take affect as expected. It only seems to be CVA which doesn't update.

I don't see any similar github issue on CVA's repo, so I assume it's something about the Shadcn config, though I haven't seen others mention this here, so I'm a bit stumped. I previously used CVA in a Next.js project, and remember sometimes having similar issues, though it was intermittent as far as I can recall.

I've followed the Shadcn guide step by step, and have even started a new project as I thought maybe my initial JS (non-TS) project was the issue, but even with a fresh TS installation of Nuxt, the result is the same. I also tried wrapping components in `<ClientOnly>` tags, but no change.

If you could suggest anything I might check to resolve this, it would be appreciated.

### Expected behavior

CVA style changes shouldn't require a server restart to take affect

_No response_

zernonia:


zernonia:
Tested the approach with super large array of component in https://github.com/zernonia/vue0/blob/main/modules/shadcn.ts

PR #252, opened 2024-01-08
fix remove today design when in the current range

valentinhttr:
Hey @sadeghbarati

Don't you think we should add a `:not(:has(.vc-day-layer))` to the selector you've added in #251 ?
```css
.calendar .vc-day.is-today:not(:has(.vc-day-layer)) {
  @apply bg-secondary rounded-md;
}
```

This avoids the problem of rounded edges when today is in the current range or the start/end of range:
CleanShot 2024-01-08 at 10 21 21
CleanShot 2024-01-08 at 10 21 44
CleanShot 2024-01-08 at 10 22 11

With the fix:
CleanShot 2024-01-08 at 10 22 56
CleanShot 2024-01-08 at 10 23 29

### DIFF
--- a/apps/www/src/lib/registry/default/ui/calendar/Calendar.vue
+++ b/apps/www/src/lib/registry/default/ui/calendar/Calendar.vue
@@ -118,7 +118,7 @@ onMounted(async () => {
 .calendar .vc-day:has(.vc-highlights) {
   @apply bg-accent first:rounded-l-md last:rounded-r-md overflow-hidden;
 }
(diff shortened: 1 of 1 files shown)


sadeghbarati:
My bad 👍

sadeghbarati:
Omg I merge it so fast 😭 

We have Two style so it's better to include the same changes in another style too (new-york)

valentinhttr:
Ohhhh yes mybad too lol! i'll create a new PR sorry

Issue #199, opened 2023-12-02
[Bug]: Some components need whitespace-nowrap


anonimusprogramus:
### Environment

```bash
Developement/Production OS: Linux Mint 20.1 Cinnamon
Node version: v18.16.0
Package manager: pnpm@8.9.2
Radix Vue version: 1.1.1
Shadcn Vue version: 0.8.2
Vue version: 3.3.7
Nuxt version: 3.8.0
Nuxt mode: universal
Nuxt target: client server
CSS framework: tailwindcss@3.3.3
Client OS: Linux Mint 20.1 Cinnamon
Browser: Brave 1.60
```

### Link to minimal reproduction

https://www.shadcn-vue.com/themes

### Steps to reproduce

1. Open [themes page](https://www.shadcn-vue.com/themes)
2. Resize to `xl` or `md` screen size
3. Reload the page
4. For miliseconds you'll see Button's title `"Copy code"` being **wrapped**

![2023-12-01 10-52-14](https://github.com/radix-vue/shadcn-vue/assets/7464749/61dfe966-9fea-44f9-af15-4a6d0abcd670)

### Describe the bug

Title having spaces may be wrapped because of missing `whitespace-nowrap` class.

In [shadcn](https://github.com/shadcn-ui/ui/tree/main/apps/www/registry/new-york/ui), following (new-york) components have `whitespace-nowrap`:
- button
- select
- tabs

Meanwhile, in [shadcn-vue](https://github.com/radix-vue/shadcn-vue/tree/dev/apps/www/src/lib/registry/new-york/ui) only these:
- table
- tabs

![2023-12-02 11-04-57](https://github.com/radix-vue/shadcn-vue/assets/7464749/caf8da51-b627-4627-937a-6831d8e67d0c)

![2023-12-02 11-05-12](https://github.com/radix-vue/shadcn-vue/assets/7464749/438d7c99-ebc9-4bbc-bcb9-c7f0cfa9965a)

### Expected behavior

No wrap

sadeghbarati:
https://github.com/radix-vue/shadcn-vue/blob/dev/apps/www/.vitepress/theme/components/PageHeaderDescription.vue#L7C1-L9C18

I think `WrapBalancer` cause the layout shifts

anonimusprogramus:
Agree about that layout shift.

[...]

PR #241, opened 2024-01-03
refactor: use `class` as prop to prevent class duplication with `cn` function

sadeghbarati:
close #35 
close #310 

### This might take a bit longer cause some of the classes/styles changed in main shadcn-ui  so I have to bring them in here too

---

Previously we used destructure `useAttrs` like 👇 , to fix class `duplication`

```ts
defineOption({
  inheritAttrs: false
})

const { class: className, ...rest } = useAttrs()
```

But destructure props or attrs will cause lose reactivity

We could make `attrs` reactive with `computed` or `toRefs`, 
But making **Always read-only (inside the component)** thing, reactive/mutable doesn't make sense 

image

---

Now we are using class in prop and delegated/rest

<details><summary>delegated/rest spread props</summary>
<p>

```vue
<script setup lang="ts">
import { type HTMLAttributes, computed } from 'vue'
import { AccordionContent, type AccordionContentProps } from 'radix-vue'
import { cn } from '@/lib/utils'

const props = defineProps<AccordionContentProps & { class?: HTMLAttributes['class'] }>()

const delegatedProps = computed(() => {
  const { class: _, ...delegated } = props

  return delegated
})
</script>

<template>
  <AccordionContent
    v-bind="delegatedProps"
    :class="
      cn(
        'overflow-hidden text-sm transition-all data-[state=closed]:animate-accordion-up data-[state=open]:animate-accordion-down',
        props.class,
      )
    "
  >
    <div class="pb-4 pt-0">
      <slot />
    </div>
  </AccordionContent>
</template>
```

</p>
</details> 

---

- After this rfcs https://github.com/vuejs/rfcs/discussions/502 comes out of **experimental** we will have this code

  <details><summary>Details</summary>
  <p>
  
  ```vue
  <script setup lang="ts">
  import type { HTMLAttributes } from 'vue'
  import { AccordionContent, type AccordionContentProps } from 'radix-vue'
  import { cn } from '@/lib/utils'
  
  const { class: className, ...delegated  } = defineProps<AccordionContentProps & { class?: HTMLAttributes['class'] }>()
  </script>
  
  <template>
    <AccordionContent
      v-bind="delegated "
      :class="
        cn(
          'overflow-hidden text-sm transition-all data-[state=closed]:animate-accordion-up data-[state=open]:animate-accordion-down',
          className,
        )
      "
    >
      <div class="pb-4 pt-0">
        <slot />
      </div>
    </AccordionContent>
  </template>
  ```
  
  </p>
  </details> 

[...]

PR #266, opened 2024-01-11
fix: add whitespace-nowrap for Button, Select, Tab

hrynevychroman:
This PR will add `whitespace-nowrap` class for Button, Select, Tab components, so there will be no overlapping when resizing, mentioned in #199 

Also in this PR I refactored Button component Props to use `VariantProps` generic instead `NonNullable<Parameters<typeof buttonVariants>>`


sadeghbarati:
I'm also working on refactoring components in this PR https://github.com/radix-vue/shadcn-vue/pull/241 which I handle button `VariantProps` in `index.ts`

Other changes are nice

hrynevychroman:
I read about `VariantProps` in issue #199 , you mentioned if someone will make PR to refactor that, but if you already do that you can dismiss that commit 🙂

sadeghbarati:
The way you handle is ok, If you don't see the Vite overlay screen while you working on the Button in localhost

---

Individual Props types in defineProps macro ✅  https://github.com/joe-bell/cva/blob/main/examples/vue/src/components/Button.vue

Entire Prop type Object in defineProps macro ❌ not yet supported in Vue

hrynevychroman:
@sadeghbarati need some advice from you, as I mentioned in [this comment](https://github.com/radix-vue/shadcn-vue/issues/199#issuecomment-1887945507) to make this we need to use tailwind custom `grid-template-columns` class `grid-cols-[minmax(0,1fr)_auto]`, but I think that it will be not good to handle it like this in code, maybe I will add some custom tailwindcss extends props, what you think?

If thou, how it will be shown, don't know how to make it simple to reuse and understand, `grid-cols-minmax-auto` maybe?

image

hrynevychroman:
In original shadcn there is flex but with `[&>span]:line-clamp-1`, I think this is worse than `truncate` because we don't see `...`, word just disappear

image

sadeghbarati:
That is a demo, user can change the styles of their component

But Yeah we can have truncate 

Line clamp is also like truncate but in Y-axis

hrynevychroman:
Yes, but what about custom TailwindCSS config grid cols, I can add them?

sadeghbarati:
Do you want to use it in Buttons Source (Registry Styles) or Demos?

hrynevychroman:
I need to use it inside `SelectTrigger.vue` component, to limit selected `span`

image

sadeghbarati:
I'm not a CSS Grid ninja but let's keep `flex` class cause width is auto by default in most elements

hrynevychroman:
Ahaha, I will try to explain this) when we have fixed block with, for example, width of 180px `Button`, inside we have 2 elements (`Text` and `Arrow`), we know that `Arrow` is fixed width, but `Text` can be different width, so when we set `display: flex;` for `Button`, when `Text` will be wider, it will move our `Arrow` further, so we need to limit `Text` to all `free` space, so we add `display: grid` to `Button` and `grid-templates-columns: minmax(0,1fr) auto;`. Minmax make all magic and limit our `Text` width, and if we add `truncate` class it will add `...` in the end of it. But with `flex` we cannot make same behaviour.

If you are assured that we can't use `grid` here, so we will make same as Original shadcn and limit by Y axis, but I think `...` is more logical 🤔 

P.S. About minmax magic I learned from official [TailwindCSS docs](https://tailwindcss.com/docs/grid-template-columns), where they use them by default for `grid-cols` classes, but main problem that predefined classes divide our block to `equal width` columns, that will not work for us 🥲

hrynevychroman:
Sorry for some misinformation, just googled some examples, and it can be fixed by `min-width: 0` 🤯 
So we don't need `grid` for this, later will change everything, `flex` works fine 🙂  

Ref: https://css-tricks.com/flexbox-truncated-text/

image

sadeghbarati:
https://twitter.com/t3dotgg/status/1745358867243171939

I'm with `min-width : 0` xd

---

First person who tweeted this was Adam Wathan


---

Thanks for the time I'll check your PR next day 
Hope you have a great day/night :cheers:

hrynevychroman:
> https://twitter.com/t3dotgg/status/1745358867243171939
> 
> I'm with `min-width : 0` xd

Just saw this in my feed 5 mins ago 😂

hrynevychroman:
@sadeghbarati PR is ready for review, added `truncate` to `SelectTrigger`, fixed `whitespace-nowrap` in `Button`, `Tabs` already has `whitespace-nowrap`

image

sadeghbarati (approved):
Thanks 🫡

PR #269, opened 2024-01-12
refactor: tailwindcss, remove all unnecessary arbitrary values, update TailwindCSS to v3.4

hrynevychroman:
**1. Remove some arbitrary values:** For example we had `translate-x-[-50%]` previously, but there is `-translate-x-1/2` predefined class 🙂 
**2. In TailwindCSS v3.4 we received extended `min/max-width/height` [predefined rules](https://tailwindcss.com/blog/tailwindcss-v3-4#extended-min-width-max-width-and-min-height-scales):** updated all values to be more native [`min-w-[8rem]` -> `min-w-32`]
**3. In TailwindCSS v3.3 we received [CSS variabled without var()](https://tailwindcss.com/blog/tailwindcss-v3-3#css-variables-without-the-var):** update all custom variables [`h-[var(--radix-select-trigger-height)]` -> `h-[--radix-select-trigger-height]`]
**4. Grid Updates.** There is some custom `grid-templates` defined, for example `grid-cols-[25px_1fr]`, but use `fr` values without wrap it inside `minmax(0,*fr)` not safe, so updated all values to be wrapped up. [`grid-cols-[25px_1fr]` -> `grid-cols-[25px_minmax(0,1fr)]`]. As a reference used default `grid-templates` rules, where `minmax` used. [Link to docs](https://tailwindcss.com/docs/grid-template-columns)

If you have some suggestions or wishes please add comments, I will update my PR ❤️


sadeghbarati:
Nice changes @romanhrynevych 💯 

We should pay more attention to the TailwindCSS blog and changelogs

All Good! can you resolve this conflicts?

hrynevychroman:
@sadeghbarati thanks 🙌 

Yeah, I will resolve conflict and notify you for review ❤️

hrynevychroman:
@sadeghbarati all conficts was in `/registry` .json files, so just rebuild it 🙂 
But there is apiToken error in Cloudflare, maybe you can help ❤️

packages/module/playground/components/ui/carousel/Carousel.vue
```diff
@@ -1,5 +1,4 @@
 <script setup lang="ts">
-import emblaCarouselVue from 'embla-carousel-vue'
```
sadeghbarati, line 2: Will Merge this PR, but it's better to ignore changes to that is not related to the PR, maybe next time 😄 

Can you also remove exact same unused imports in `Carousel.vue` **default** and **new-york** styles too?

Thanks

hrynevychroman: sorry, to resolve confict I just merge `dev` into my branch to resolve conflicts locally, how can I do it better next time?

sadeghbarati: Ow yes you right ❤️  my bad

apps/www/src/examples/cards/components/GitHubCard.vue
```diff
@@ -26,7 +26,7 @@ import { Separator } from '@/lib/registry/new-york/ui/separator'
 
 <template>
   <Card>
-    <CardHeader class="grid grid-cols-[1fr_110px] items-start gap-4 space-y-0">
+    <CardHeader class="grid grid-cols-[minmax(0,1fr)_110px] items-start gap-4 space-y-0">
```
sadeghbarati, line 29: If TailwindCSS add this 👇  style to their `reset` file

```css
* {
min-width: 0;
}
```

We can remove `minmax`?

sadeghbarati:
> But there is apiToken error in Cloudflare

Don't worry about it Zernonia will fixed it later

apps/www/src/examples/cards/components/GitHubCard.vue
hrynevychroman: Yes, it will work fine, made some reproduction [codepen](https://codepen.io/greenwich/pen/oNVYpJm), just remove css and see that grid layout would shift from `container` block 🙂 

Maybe we will add this styles by default, or will wait TailwindCSS v3.5?

sadeghbarati: Perfect.

Rebuild registry again then we can merge this 🎉

hrynevychroman: So, we add this 👇 to our default styles? If yes, where I can find it, because I don't know codebase perfectly 😁 
```css
* {
    min-width: 0;
}
```

sadeghbarati: @zernonia Sry for taging you too much

https://www.joshwcomeau.com/css/interactive-guide-to-flexbox/#the-minimum-size-gotcha-11

I can't decide of this change can you check if it's worth to add or we should wait until
TailwindCSS add `min-width: 0` change to their reset.css

---

> where I can find it

You can find it here

`apps/www/scripts/build-registry.ts` -> `export const BASE_STYLES_WITH_VARIABLES`

hrynevychroman: We will wait for Zernonia approve and then merge it 🫂

valentinhttr:
Well done @romanhrynevych for this update !
As a small note, what do you think of using the `has-[]` variant [that appeared in version 3.4](https://tailwindcss.com/blog/tailwindcss-v3-4#new-has-variant) aswell?

I know that `has()` is widely used in the calendar component, for example. I don't know if that would complicate things more than it would simplify them, what do you think?

hrynevychroman:
> Well done @romanhrynevych for this update ! As a small note, what do you think of using the `has-[]` variant [that appeared in version 3.4](https://tailwindcss.com/blog/tailwindcss-v3-4#new-has-variant) aswell?
> 
> I know that `has()` is widely used in the calendar component, for example. I don't know if that would complicate things more than it would simplify them, what do you think?

I think there is no need for making that changes, in Calendar.vue used <style> tag with @apply styles, if styles was defined like classNames maybe, but now I think we can just leave this)

hrynevychroman:
Also, I'm not a very big fan of `@apply` styles, because than tailwind.css file becomes bigger. For example if you have `w-full` class in 100 places of code, it will only add once, but if you assign it by `@apply`, tailwindcss must write new class selector and add this styles, so all magic of tailwindcss is just disappeared 🥲

packages/module/playground/components/ui/carousel/Carousel.vue
zernonia: ahh I added this import else the registry dependencies can't track it well 😂
but yeah it's alright. I will update the build registry script properly 😂

apps/www/src/examples/cards/components/GitHubCard.vue
zernonia: Thanks for the awesome job guys! I'm looking at it now 😁

zernonia: Let's be safe for now and keep `minmax`. As we are not sure if and when TW will include min-width in their reset. 😁

sadeghbarati (approved):

sadeghbarati:
Dam, I'll resolve the conflict

PR #272, opened 2024-01-14
fix: select popper position

### LINKED ISSUE #245: [Bug]: Select dropdown list does not auto-position depending on the edge of the screen
### Environment

```bash
Developement/Production OS: Windows 10 19043.1110
Node version: 16.0.0
Package manager: pnpm@8.6.0
Radix Vue version: 1.0.0
Shadcn Vue version: 1.0.0
Vue version: 3.0.0
Nuxt version: 3.0.0
Nuxt mode: universal
Nuxt target: server
CSS framework: tailwindcss@3.3.3
Client OS: Windows 10 19043.1110
Browser: Chrome 90.0.4430.212
```

none

### Steps to reproduce

Go to https://www.shadcn-vue.com/examples/cards.html, scroll down to the Payment Method card, click on Select.

### Describe the bug

Dropdown list does not auto-position depending on the edge of the screen.

### Expected behavior

The dropdown list should auto-position depending on the edge of the screen.

image

antonreshetov:
Fix #245

### DIFF
--- a/apps/www/src/lib/registry/default/ui/select/SelectContent.vue
+++ b/apps/www/src/lib/registry/default/ui/select/SelectContent.vue
@@ -13,6 +13,7 @@ const props = withDefaults(
   defineProps<SelectContentProps & { class?: string }>(), {
     position: 'popper',
     sideOffset: 4,
+    avoidCollisions: true,
   },
 )
 const emits = defineEmits<SelectContentEmits>()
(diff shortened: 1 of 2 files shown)


sadeghbarati:
Wonder why default **avoidCollisions: true** `radix-vue` prop is not affecting poppers in `shadcn-vue`

https://github.com/radix-vue/radix-vue/blob/main/packages/radix-vue/src/Popper/PopperContent.vue#L20

antonreshetov:
@sadeghbarati When can we expect the release with this fix?

PR #274, opened 2024-01-15
fix: prevent page zoom while tapping carousel buttons

### LINKED ISSUE #273: [Bug]: When tapping the previous/next buttons multiple times it triggers page zoom
Hello from the Embla side, I hope you're all doing well 👋!

### Environment

```bash
Device: iPhone 8
Browser: Safari iOS 16.7.4
```

https://www.shadcn-vue.com/docs/components/carousel.html

### Steps to reproduce

- Browse to the [carousel page](https://www.shadcn-vue.com/docs/components/carousel.html) on an iOS device like an iPhone.
- Tap on the next button twice or more to scroll through the carousel slides.
- See the page gets zoomed which probably isn't what you wanted to do.

### Describe the bug

When tapping the carousel previous and next buttons multiple times on an iOS device, it sometimes triggers an unwanted page zoom.

### Expected behavior

Tapping the carousel previous and next buttons should only trigger carousel navigation and not page zoom.

## Suggested solution
I believe adding `touch-action: manipulation;` [here](https://github.com/radix-vue/shadcn-vue/blob/dev/apps/www/src/lib/registry/default/ui/carousel/CarouselPrevious.vue#L16) and [here](https://github.com/radix-vue/shadcn-vue/blob/dev/apps/www/src/lib/registry/default/ui/carousel/CarouselNext.vue#L16) solves the problem. As [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/touch-action#manipulation) describes it:

> Enable panning and pinch zoom gestures, but disable additional non-standard gestures such as double-tap to zoom. Disabling double-tap to zoom removes the need for browsers to delay the generation of click events when the user taps the screen.

Thank you for your time!

sadeghbarati:
close https://github.com/radix-vue/shadcn-vue/issues/273

Thanks, David for the report ❤️

### DIFF
--- a/apps/www/.vitepress/theme/components/ComponentPreview.vue
+++ b/apps/www/.vitepress/theme/components/ComponentPreview.vue
@@ -52,7 +52,7 @@ const { style } = useConfigStore()
           </div>
         </div>
         <div
-          :class="cn('preview flex min-h-[350px] w-full justify-center p-6 lg:p-10', {
+          :class="cn('preview flex min-h-[350px] w-full justify-center p-10 items-center', {
             'items-center': align === 'center',
             'items-start': align === 'start',
             'items-end': align === 'end',
--- a/apps/www/src/content/docs/components/carousel.md
+++ b/apps/www/src/content/docs/components/carousel.md
@@ -48,18 +48,6 @@ import {
 
 ## Examples
 
-### Orientation
-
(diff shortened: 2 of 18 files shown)


sadeghbarati:
Hi @zernonia 👋 

Request Deploy is still failing

PR #278, opened 2024-01-16
chore: update deps and vitepress

sadeghbarati:
- Vitepress now uses `Shikiji` instead of `Shiki` which not has a built-in `css-variables` theme

- Also updated `tailwindcss` cause of type error in config.mts (vitepress config) which I could not cherry-pick from Roman Hrynevych's PR


zernonia (approved):
lgtm! and finally "Request deploy" works now!

Issue #158, opened 2023-11-10
[Feature]: Make it possible to use CommandInput with our own filtering method


maelp:
### Describe the feature

I want to do a "CommandDialog" with a CommandInput but use Algolia to search in my database collections.

<img width="543" alt="image" src="https://github.com/radix-vue/shadcn-vue/assets/15458/f37ff080-861c-42ac-9e79-3cee93ee88bd">

The issue I have is that if I use CommandInput, it automatically does the filtering using each CommandItem "value" but I want to avoid this since I create the CommandItem dynamically with the algolia search results, and the algolia filters are more complex than just filtering text on "value"

also there is a weird issue if I change the filters by typing, deleting, starting again, the "CommandItem" seem to be accessed in a random order (and some are not accessible) when using up and down keys? (you will see the selected item "jump around" sometimes, although I only press either "down key" multiple times, or "up key" multiple times, but I don't mix them)

https://github.com/radix-vue/shadcn-vue/assets/15458/fee362d0-a38d-409d-81b6-c70db90d430e

### Additional information

- [ ] I intend to submit a PR for this feature.
- [ ] I have already implemented and/or tested this feature.

maelp:
To make it clear: I'd like a way to do the CommandDialog where the CommandInput only lets me select options with up/down, but does not filter on its own the content, so that I can use something like this to dynamically replace the result, and still use the "up/down/enter" selection

```
  <CommandDialog :open="open" @update:open="(v) => (open = v)">
    <ais-instant-search
      :search-client="searchClient"
      :index-name="searchIndexes.batteryAssemblies"
      :stalled-search-delay="200"
    >
      <ais-configure :hits-per-page.camel="5" />
      <ais-search-box placeholder="Type to search...">
        <template v-slot="{ refine }">
          <CommandInput
            @input="refine($event.currentTarget.value)"
            placeholder="Type to search..."
          ></CommandInput>
        </template>
      </ais-search-box>
      <algolia-loading-indicator />
      <CommandList>
        <CommandEmpty>No results found.</CommandEmpty>
        <CommandGroup
          :heading="category.title"
          v-for="category in categories"
          :key="category.value"
        >
          <ais-index :index-name="category.searchIndex">
            <ais-hits>
              <template #item="{ item }">
                <CommandItem
                  :value="item.id"
                  :key="item.id"
                  @click="
                    (event) => openCategoryId(event, category.value, item.id)
                  "
                >
                  {{ item.id }}
[...]

PR #284, opened 2024-01-17
[Feature]: Make custom filter-function for CommandDialog Component

hrynevychroman:
Main reason of this PR is to close this issue #158 , but think that it will be helpful to add new Demo with `@vueuse/useFuse` filtering

Have some errors with creating new Demo, @sadeghbarati need help from you because previously I never worked with this type of projects 🙂 

image

Link to Demo: `www/src/lib/registry/default/example/CommandDialogUseFuseDemo.vue`

Styling and other will fixed later, need to understand how to add a new Demo code 😁


sadeghbarati:
I'm on it, will explain after commit

hrynevychroman:
Thanks!

sadeghbarati:
When you have create demos you have two options 

1. update `apps/www/__registry__/index.ts` manually (add demos)

2. you can run `pnpm build:registry` when you created demos in both styles

hrynevychroman:
Great, thanks for explanation! ❤️ 

Need some help with this issue, because it contains many issues inside 😁  Now I added custom `filter-function` props to `CommandDialog.vue` component, but there are some other suggestion and I don't know if we need to fix them)

1. In Issue #158 you can see example with `Algolia` (API Fetch) `CommandDialog`, but there is some problems with dynamically change `CommandItems` inside `CommandList`, don't think that after change keyboard navigation will work at all 🙂 
2. There is some bug with `useFuse` too, if I add custom filter that will return output with shuffled order, keyboard navigation don't work properly, for fix I added `shouldSort: false` to `useFuse.options.fuseOptions`, but think that this must be fixed too

So one bug that need to be fixed now, dynamic change of `CommandItem` (insert, delete) that will toggle keyboard navigation logic to refresh. I'm not familiar with `radix-vue` on this level, maybe someone can help here.

I will try to fix this, need some help / time 🙂

sadeghbarati:
No need to rush 👍 

There is also the same pkg for Vue called [vue-command-palette](https://github.com/xiaoluoboding/vue-command-palette) which uses `Fuse`, you can take a look at it and see how it handles the search/sort/filtering

sadeghbarati:
@zernonia can we remove these aggressive warnings for a11y?

Other shadcns seem not to have these logs


<details><summary>Image of logs</summary>
<p>

image

</p>
</details> 


---

Also, there is something wrong with `CommandDialog` which not showing `DialogOverlay`

https://stackblitz.com/edit/5h3b5u?file=src%2FApp.vue

maelp:
Will this be merged?

This is how I'm solving it in CommandDialog for now

```
const props = defineProps<
  DialogRootProps & {
    filterFunction?: (val: Array<string | any>, term: string) => Array<any>;
  }
>();
const emits = defineEmits<DialogRootEmits>();

const forwarded = useForwardPropsEmits(props, emits);
const forwardedExceptFilterFunction = computed(() => {
  const { filterFunction: _, ...rest } = forwarded.value;
  return rest;
});
```

apps/www/src/lib/registry/default/ui/command/CommandDialog.vue
```diff
@@ -1,19 +1,31 @@
...
+})
+
+const forwarded = useForwardPropsEmits(delegatedProps, emits)
 </script>
 
 <template>
   <Dialog v-bind="forwarded">
     <DialogContent class="overflow-hidden p-0 shadow-lg">
-      <Command class="[&_[cmdk-group-heading]]:px-2 [&_[cmdk-group-heading]]:font-medium [&_[cmdk-group-heading]]:text-muted-foreground [&_[cmdk-group]:not([hidden])_~[cmdk-group]]:pt-0 [&_[cmdk-group]]:px-2 [&_[cmdk-input-wrapper]_svg]:h-5 [&_[cmdk-input-wrapper]_svg]:w-5 [&_[cmdk-input]]:h-12 [&_[cmdk-item]]:px-2 [&_[cmdk-item]]:py-3 [&_[cmdk-item]_svg]:h-5 [&_[cmdk-item]_svg]:w-5">
+      <Command :filter-function="props.filterFunction" class="[&_[cmdk-group-heading]]:px-2 [&_[cmdk-group-heading]]:font-medium [&_[cmdk-group-heading]]:text-muted-foreground [&_[cmdk-group]:not([hidden])_~[cmdk-group]]:pt-0 [&_[cmdk-group]]:px-2 [&_[cmdk-input-wrapper]_svg]:h-5 [&_[cmdk-input-wrapper]_svg]:w-5 [&_[cmdk-input]]:h-12 [&_[cmdk-item]]:px-2 [&_[cmdk-item]]:py-3 [&_[cmdk-item]_svg]:h-5 [&_[cmdk-item]_svg]:w-5">
```
sadeghbarati, line 28: @romanhrynevych can you check this line, it has TypeScript error

maelp:
if it can help, for me when I do Algolia search I do the following:

- I bind the CommandDialog update:input to a method which triggers a query to algolia and get the results in a list
- I show the list (without filering, eg `filtering-function = (list, term) => list`

so each time I type, I let Algolia do the "filtering" in the backend and return the result, and I display them "without filtering" (from CommandDialog) so that the users sees exactly the results of algolia

sadeghbarati:
Let's close stale PRs xd

filterFunction is also going to remove in v2

PR #309, opened 2024-01-27
[Bug Fix] ScrollBar width

PiervitoCreanza:
Hi,
I think I found a bug in the ScrollBar component.
In the image below, notice that the `flex-1` class assigned to the `ScrollAreaThumb` causes the scrollbar to expand and fill the entire width of the scroll area.
immagine

Removing the specified class produces the expected behaviour:
immagine

### DIFF
--- a/apps/www/src/lib/registry/default/ui/scroll-area/ScrollBar.vue
+++ b/apps/www/src/lib/registry/default/ui/scroll-area/ScrollBar.vue
@@ -18,6 +18,6 @@ const props = withDefaults(defineProps<ScrollAreaScrollbarProps>(), {
            && 'h-2.5 border-t border-t-transparent p-[1px]',
          $attrs.class ?? '')"
   >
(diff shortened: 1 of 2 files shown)


sadeghbarati:
Hi @PiervitoCreanza 

- shadcn-ui: 
https://github.com/shadcn-ui/ui/blob/main/apps/www/registry/default/ui/scroll-area.tsx#L38

- shadcn-vue: 
https://github.com/radix-vue/shadcn-vue/blob/dev/apps/www/src/lib/registry/default/ui/scroll-area/ScrollBar.vue#L18

Looks like we are `missing` **flex-col** in orientation === "horizontal"

Can you turn back **flex-1** and add **flex-col** to the `ScrollAreaScrollbar` Component

PiervitoCreanza:
Oh, you're right. I reverted my old commits and pushed the new one.

PR #319, opened 2024-02-03
fix(cli): add `xl` item to borderRadius in tailwind.config template

itsTPM:
already did this in #231, but seems like a829212d4283d9312d6948bd979366706cf2b72d undid that

### DIFF
--- a/apps/www/src/lib/registry/new-york/ui/card/Card.vue
+++ b/apps/www/src/lib/registry/new-york/ui/card/Card.vue
@@ -11,7 +11,7 @@ const props = defineProps<{
   <div
     :class="
       cn(
-        'rounded-xl border bg-card text-card-foreground shadow',
+        'rounded-lg border bg-card text-card-foreground shadow',
         props.class,
       )
     "
--- a/apps/www/src/lib/registry/new-york/ui/card/Card.vue
+++ b/apps/www/src/lib/registry/new-york/ui/card/Card.vue
@@ -11,7 +11,7 @@ const props = defineProps<{
   <div
     :class="
       cn(
-        'rounded-lg border bg-card text-card-foreground shadow',
+        'rounded-xl border bg-card text-card-foreground shadow',
         props.class,
       )
     "
--- a/apps/www/src/public/registry/styles/new-york/card.json
+++ b/apps/www/src/public/registry/styles/new-york/card.json
@@ -7,7 +7,7 @@
   "files": [
     {
       "name": "Card.vue",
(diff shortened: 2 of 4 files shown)


sadeghbarati:
Sry that was a big refactor I did not notice that


not my quote `shadcn-svelte` issue section
> This project's goal is to remain aligned with [shadcn/ui](https://github.com/shadcn-ui/ui). If the feature you are requesting is not in the original shadcn/ui, it will not be considered here.


https://github.com/shadcn-ui/ui/blob/main/apps/www/registry/new-york/ui/card.tsx#L12

https://github.com/huntabyte/shadcn-svelte/blob/main/apps/www/src/lib/registry/new-york/ui/card/card.svelte#L13

sadeghbarati:
I'll change those Card to use `default` style to keep the radius from Customize popup sound good?

itsTPM:
customization popup changes the radius for the rounded-lg class, which is what the cards in shadcn-vue have (as i noticed), but you ofc can use a better solution, if there is such a thing

itsTPM:
> Sry that was a big refactor I did not notice that
> 
> not my quote `shadcn-svelte` issue section
> 
> > This project's goal is to remain aligned with [shadcn/ui](https://github.com/shadcn-ui/ui). If the feature you are requesting is not in the original shadcn/ui, it will not be considered here.
> 
> https://github.com/shadcn-ui/ui/blob/main/apps/www/registry/new-york/ui/card.tsx#L12
> 
> https://github.com/huntabyte/shadcn-svelte/blob/main/apps/www/src/lib/registry/new-york/ui/card/card.svelte#L13

this issue seems to exist in shadcn-svelte as well, if shadcn-svelte also uses `rounded-ld` class for customizing border-radius (or for react shadcn too, i'll check this later)

sadeghbarati:
@itsTPM

I think we must add another option to our tailwind.config `xl` 

https://github.com/shadcn-ui/ui/blob/main/tailwind.config.cjs#L52

You can add it here 👇 
https://github.com/radix-vue/shadcn-vue/blob/dev/packages/cli/src/utils/templates.ts#L119

itsTPM:
yes, that's the point
thanks!

sadeghbarati:
@itsTPM since you are in `dev` branch I can't checkout to yours cause I already have dev branch

Can you add same thing to this file in `apps/www`, Thanks 👍 
https://github.com/radix-vue/shadcn-vue/blob/dev/apps/www/tailwind.config.js#L52

sadeghbarati:
Will fix [themes route cards](https://www.shadcn-vue.com/themes.html) after this, Thanks for bringing it up

PR #325, opened 2024-02-05
feat: pin input

sadeghbarati:


apps/www/src/lib/registry/new-york/example/PinInputFormDemo.vue
```diff
@@ -0,0 +1,78 @@
...
+  FormDescription,
+  FormField,
+  FormItem,
+  FormLabel,
+  FormMessage,
+} from '@/lib/registry/new-york/ui/form'
+import { toast } from '@/lib/registry/new-york/ui/toast'
+
+const formSchema = toTypedSchema(z.object({
+  pin: z.array(z.string().min(1)).length(5),
```
sadeghbarati: not sure about this `zod` schema

PR #334, opened 2024-02-13
feat: implement new `Tasks BigData` example for useVirtualList usage

### LINKED ISSUE #296: [Bug]: DropdownMenu(Trigger) will be slow to display when there's a lot of render elements
### Environment

```bash
Shadcn Vue version: 0.8.7
Vue version: 3.4.15
```

https://github.com/frankchen211/shadcn-vue/tree/draft/blue-violet

### Steps to reproduce

[For task page example](https://www.shadcn-vue.com/examples/tasks.html), 
1. Make more data like 1000 rows
2. Set pageSize to 1000
3. click `View` button to open the dropdown menu, it will take a while to display(remove any functions from [TanStack Table](https://tanstack.com/table/v8)).
```
const columns = ['col1', 'col2', 'col3']
```
```
<DropdownMenuCheckboxItem
        v-for="column in columns"
        :key="column"
        class="capitalize"
        :checked="true"
>
        {{ column }}
</DropdownMenuCheckboxItem>
```
4. click and compare with the FacetedFilter(using Popover) which is faster.

### Describe the bug

[For task page example](https://www.shadcn-vue.com/examples/tasks.html), if you have large dataset to display at one page,
like 1000 per/page, then click `View` button to open the dropdown menu, it will take a while to display, but the FacetedFilter(using Popover) is faster.

### Expected behavior

Faster

_No response_

hrynevychroman:
This PR try to fix #296 

So I add a new example where implemented `virtualList` based on vueuse core function.

I tried to use seperate package but it don't fit for `table` HTML at all because creating 2 div wrappers. 

Please watch this video 👇 , or try by yourself, you will feel performance improvements at all

https://github.com/radix-vue/shadcn-vue/assets/82209198/50fa452d-a92a-4c6f-a950-963b27d77e38

@sadeghbarati need help with this, what will be the best Doc for this, because if we are porting only original package, there is no `BigData` example at all 🙂


sadeghbarati:
Thanks, Man! @romanhrynevych 

Will try to checkout to your branch to fully test it ❤️ 

Also, I think the issue you linked is related to `DropdownMenu` itself and its `DropdownMenuItems`
It might be related to this discussion is [Discord Help and Question (Combobox or Dropdown Virtual List)](https://discord.com/channels/1121805548782878813/1192153099603816540) but thanks anyway 😁 

---

> need help with this, what will be the best Doc for this, because if we are porting only original package, there is no BigData example at all

Can you explain more please 🙏

apps/www/.vitepress/config.mts
```diff
@@ -70,6 +70,13 @@ export default defineConfig({
   rewrites: {
     'content/(.*)': '(.*)',
   },
+  vue: {
+    template: {
+      compilerOptions: {
+        isCustomElement: tag => tag === 'option.icon',
+      },
```
sadeghbarati: I think there is no need for it cause we don't use `Web Components` in this project

If you want to use `option.icon` use it like

```html
<component :is="option.icon" />
```

https://vuejs.org/guide/essentials/component-basics#dynamic-components

hrynevychroman:
> > need help with this, what will be the best Doc for this, because if we are porting only original package, there is no BigData example at all
> 
> 
> 
> Can you explain more please 🙏 

Don't think that this place on main page is great position for example, maybe you will move it to another place 😁❤️

apps/www/.vitepress/config.mts
hrynevychroman: I was receiving vue/core warning in devtools console, so added this lines for IDE and vue/core fix ☺️

sadeghbarati:
> Don't think that this place on main page is great position for example, maybe you will move it to another place 😁❤️

~~I think it's better to replace it with the Original Task page~~ let see what is Zernonia's opinion is

I think in `data-table.md` after column pinning example is good place

hrynevychroman:
@zernonia can you take a look on this PR? What you think is the best place to show Tasks Example?

apps/www/.vitepress/config.mts
zernonia: Yes. Change `<option.icon>` to `<component :is="option.icon" />` will render the icon nicely. We dont need custom element here.

zernonia:
Thanks for the PR @romanhrynevych . I don't think we need to duplicate a whole other component just to showcase large dataset. Most of the component/utilities are the same and can be reuse from `examples/tasks`.

And let's create another path (https://www.shadcn-vue.com/docs/guide/rendering-large-data.html), and write some guide/steps to help other dev to use the virtualisation package instead of another demo.

I believe this way it will be more clear, and more helpful.

apps/www/.vitepress/config.mts
hrynevychroman: Changed in this commit https://github.com/radix-vue/shadcn-vue/pull/334/commits/b09910fb2eb6f6dbad20a4cb6971cb6f17f5f4c7

hrynevychroman:
Ok, I will update this PR to your advice soon ❤️

hrynevychroman:
@sadeghbarati hi 👋 

I added new `guide` section, but don't know where to move `examples/big-data` folder, because it has custom logic for this type of tables only, maybe you will provide nice place for it ❤️ 

If it will be some more fixes feel free to provide them, i will fix all of them 😁

sadeghbarati:
Thanks Roman will check

hrynevychroman:
@sadeghbarati just noticed that tanstack has his own virtualizer (noticed in this Radix-vue Issue your comment 😁 ) https://github.com/radix-vue/radix-vue/issues/696#issuecomment-1953807473, maybe here I will replace vueuse integration with tanstack table example?

Link: https://tanstack.com/virtual/latest/docs/framework/vue/examples/table

sadeghbarati:
Feel free to switch to TanStack Virtual or add both?

hrynevychroman:
thinking about switching to TanStack for more extended features in future 🤔

sadeghbarati:
Yes TanStack for the Win

maelp:
I've found tanstack virtualized more difficult to use (and finnicky) than vue-use, but vue-use has less features

Coudln't really manage to make this work with a table without hitting display issues, so I ended up adding fixed-width to all cells and using a "list" with each row made of cells

PR #336, opened 2024-02-13
fix: update SelectItem.vue to reduce left-padding

### LINKED ISSUE #335: [Bug]: The SelectItem has the wrong left-padding class
### Reproduction

https://www.shadcn-vue.com/docs/components/select.html

### Describe the bug

It seems the "default" style of component has the wrong left padding class ("pl-8") resulting in a big gap to the left of each item in the select:

https://github.com/radix-vue/shadcn-vue/blob/41660d36cdb0d4eb8b22ab1efe815d6a5c1fb4a9/apps/www/src/lib/registry/default/ui/select/SelectItem.vue#L29

I believe that _should_ be "pl-2", like it is for the "new york" style:

https://github.com/radix-vue/shadcn-vue/blob/41660d36cdb0d4eb8b22ab1efe815d6a5c1fb4a9/apps/www/src/lib/registry/new-york/ui/select/SelectItem.vue#L29

### System Info

```bash
System:
 OS: macOS 14.1.1
Binaries:
 Node: 20.11.0 - /usr/local/bin/node
 npm: 10.2.4 - /usr/local/bin/npm
 bun: 1.0.7 - ~/.bun/bin/bun
Browsers:
 Chrome: 121.0.6167.184
 Safari: 17.1
```

### Contributes

- [ ] I am willing to submit a PR to fix this issue

gregmsanderson:
This should fix https://github.com/radix-vue/shadcn-vue/issues/335

### DIFF
--- a/apps/www/src/lib/registry/default/ui/select/SelectItem.vue
+++ b/apps/www/src/lib/registry/default/ui/select/SelectItem.vue
@@ -26,7 +26,7 @@ const forwardedProps = useForwardProps(delegatedProps)
     v-bind="forwardedProps"
     :class="
       cn(
-        'relative flex w-full cursor-default select-none items-center rounded-sm py-1.5 pl-8 pr-2 text-sm outline-none focus:bg-accent focus:text-accent-foreground data-[disabled]:pointer-events-none data-[disabled]:opacity-50',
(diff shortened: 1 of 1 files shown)


gregmsanderson:
Closed: will instead move the check to the left (meaning `p-8` will still be needed)

PR #339, opened 2024-02-14
fix: add `Geist` default font

### LINKED ISSUE #332: [Feature]: Add a default font for shadcn-vue.com
### Describe the feature

Thanks for the great project. Both shadcn.com and shadcn-svelte.com used Geist as the default font. I think shadcn-vue should also have one instead of system sans serif to make it a consistent experience on different devices. Currently it doesn't look nice on my device using YaHei as the system font.

image

### Additional information

hrynevychroman:
This PR fix #332 

@sadeghbarati please watch on eslint config, i write to you in discord ❤️


sadeghbarati:
Nice choice 👍

**Variable Fonts** >>>>

sadeghbarati:
I think `.woff2` is enough/better cause `.ttf` is for older browsers which is not important

Just moved variable font to public/static directory and add `LICENSE.TXT` just like `shadcn-svelte`

I'll create new PR for eslint and documentation arrangements

PR #347, opened 2024-02-15
chore: update antfu config

sadeghbarati:
- Most of major updates in root `package.json` was dropping node 16 so we can update them without worrying about breaking changes

- Removed `pnpm` from `devDependencies` cause it's unnecessary when we have it in our global environment path

Did not run `eslint` cause it might bring conflicts to other PR let's merge them all then run it for rest of the project


MuhammadM1998:
May I ask why you're using taze instead of renovate? I use taze locally but renovate is better for repos as it automatically checks for updates, creates the pr etc

sadeghbarati:
@MuhammadM1998 Hi, sure!

I agree with having GitHub bots take care of updating dependencies but the noise is annoying

If we can control the GitHub bot update schedules I glad to have it in this project, but I'm not the one who can add this bot

MuhammadM1998:
Thanks for replying! The schedules are configurable and you can set it up to auto group updates in one PR to reduce the clutter even more.

PR #367, opened 2024-03-01
Fix: form capitalization

braden-w:
`Form` should be cased as `form` to match the starting and closing tags and be consistent with docs.


sadeghbarati:
Thanks

PR #368, opened 2024-03-01
Update CommandDialog.vue to add custom filter-function

maelp:
Add a custom filter-function to CommandDialog

### DIFF
--- a/apps/www/src/lib/registry/default/ui/command/CommandDialog.vue
+++ b/apps/www/src/lib/registry/default/ui/command/CommandDialog.vue
@@ -4,16 +4,24 @@ import type { DialogRootEmits, DialogRootProps } from 'radix-vue'
 import Command from './Command.vue'
 import { Dialog, DialogContent } from '@/lib/registry/default/ui/dialog'
 
-const props = defineProps<DialogRootProps>()
+const props = defineProps<
+  DialogRootProps & {
+    filterFunction?: (val: Array<string | any>, term: string) => Array<any>;
+  }
+>();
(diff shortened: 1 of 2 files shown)


hrynevychroman (approved):
Seems good)

sadeghbarati:
@maelp can you add exact same changes to `new-york` style?

sadeghbarati:
@romanhrynevych 

What is the difference between this PR and your PR in https://github.com/radix-vue/shadcn-vue/pull/284

hrynevychroman:
@sadeghbarati I was trying to fix that TS error, but nothing helped at all, I see that in this PR maelp destructure filter function and provide it as props to `Command` component

Also, I don't think that it is a great idea to overlap default `radix-vue` props interface for Dialog, maybe change firstly it inside radix-vue and then test here?

This is type of errors depends on `radix-vue` codebase, not `shadcn-ui` at all 😢

maelp:
@sadeghbarati I did the change for NY

sadeghbarati:
I'm still not sure about this PR

The current Command (Combobox) in `radix-vue` has some problems adding a props to add another custom filter function is good? Idk

Maybe share a video (with open devtools) here to see what PR is about

---

https://github.com/xiaoluoboding/vue-command-palette

I'll try to make PR to `vue-command-palette` to update their deps then try port the pkg to `shadcn-vue` until `radix-vue` Command become ready

sadeghbarati:
Let's wait for radix-vue `Listbox` implementation it should fix Combobox issues

PR #377, opened 2024-03-04
docs: add new example for slider in form

Knogobert:
Saw that we didn’t have a nice example for Sliders usage in forms. This should make it more obvious that support is there.


sadeghbarati:
I forgot branch must be upstream to run deployment in Cloudflare pages

Thanks btw!

Knogobert:
Nice!
You're one quick boi.

I just realized that this could probably be good to have on the other shadcn repos. Or how tied are they all?

Issue #378, opened 2024-03-04
[Feature]: Preprocess and host JS components


WarningImHack3r:
### Describe the feature

Similar to what has been made on shadcn-svelte (https://github.com/huntabyte/shadcn-svelte/pull/715), it would be great to create a "server-hosted" version of the components, available from a registry endpoint (`/registry/styles/xxx-js/component.json`).

This would help for 2 things: avoid additional client-side computing for each JS component downloaded, as well as unlock that JS-components feature for [my shadcn IntelliJ plugin](https://plugins.jetbrains.com/plugin/23479-shadcn-ui-components-manager). (Transpiling Vue TS files from a Java-like codebase is not an easy task, and i would prefer avoiding that heavy task for the Vue shadcn port only)

It shouldn’t require much work here as it’s pretty much a copy-paste from the Svelte version. What do you think about it?

### Additional information

- [ ] I intend to submit a PR for this feature.
- [ ] I have already implemented and/or tested this feature.

zernonia:
Instead of adding yet another sets of `registry`, I think we can try a different approach. I'm thinking of a `js` branch, that would deploy already transformed `js` registry. 

When `ci` detects new changes to `registry` files in the `dev` branch, it will create a PR with the updated `js` code. 
We can simply updates which registry url to hit for the CLI based on user config.

WDYT @sadeghbarati ?

sadeghbarati:
That would be great 💯

WarningImHack3r:
> Instead of adding yet another sets of `registry`, I think we can try a different approach.

Why creating a branch/(auto-)PR mechanism when you could simply update the `build_registry.ts` file? Seems a lot of work for a same end to me 🤔

zernonia:
The reason im suggesting above is that there's tooooo many files inside the current repo now. It makes navigation harder, and might put heavy loads on vscode when developing. 😂

WarningImHack3r:
> The reason im suggesting above is that there's tooooo many files inside the current repo now.

As long as it ends up making JS files available through an endpoint, I'm fine with any method!

zernonia:
Decided not to go with this approach as we want to support for other registry such as [Inspria UI](https://inspira-ui.com/), and needing the registry author to parse and perform `detype` transformation is just another unnecessasry step.

PR #379, opened 2024-03-05
fix: change to css logical props

### LINKED ISSUE #357: [Feature]: Support RTL in shadcn components
### Describe the feature

It seems that components use left and right values from tailwindcss 
i would recommend using start and end so it supports RTL and LTR directions in projects

so instead of `pr-1` , `ml-2` ...
we use `pe-1`, `ms-2`
And so on 

also in components that have Chevron icons 
we should add `rtl:rotate-180` class so we make sure when we change direction arrow rotate

Tasks:
- [ ]  use ( `start-, end-` / `ps-, pe-` / `ms-, me- `/ ... ) instead of (`left-, right-` / `pl-, pr-` / `ml-, mr-` / ... )
- [ ]  add rotation transformation to chevrons that point to direction 
- [ ]  related bug https://github.com/radix-vue/radix-vue/issues/719

### Additional information

lord007tn:
I did change to css logical props
add rotation in rtl from chevrons and arrows 
except the one in the onKeyDown function in carousel as it require a testing on what direction the website is
i belive this will change after we settle on the RTL/LTR provider for components

i changed text to be on start rather than forced to be left as it's more meaningful to what direction the text is written

### DIFF
--- a/apps/www/src/lib/registry/default/ui/context-menu/ContextMenuItem.vue
+++ b/apps/www/src/lib/registry/default/ui/context-menu/ContextMenuItem.vue
@@ -8,7 +8,7 @@ import {
 } from 'radix-vue'
 import { cn } from '@/lib/utils'
 
-const props = defineProps<ContextMenuItemProps & { class?: HTMLAttributes['class']; inset?: boolean }>()
+const props = defineProps<ContextMenuItemProps & { class?: HTMLAttributes['class'], inset?: boolean }>()
 const emits = defineEmits<ContextMenuItemEmits>()
 
 const delegatedProps = computed(() => {
@@ -25,7 +25,7 @@ const forwarded = useForwardPropsEmits(delegatedProps, emits)
     v-bind="forwarded"
     :class="cn(
       'relative flex cursor-default select-none items-center rounded-sm px-2 py-1.5 text-sm outline-none focus:bg-accent focus:text-accent-foreground data-[disabled]:pointer-events-none data-[disabled]:opacity-50',
-      inset && 'pl-8',
+      inset && 'ps-8',
       props.class,
     )"
   >
--- a/apps/www/src/lib/registry/default/ui/context-menu/ContextMenuLabel.vue
+++ b/apps/www/src/lib/registry/default/ui/context-menu/ContextMenuLabel.vue
@@ -3,7 +3,7 @@ import { type HTMLAttributes, computed } from 'vue'
 import { ContextMenuLabel, type ContextMenuLabelProps } from 'radix-vue'
 import { cn } from '@/lib/utils'
 
-const props = defineProps<ContextMenuLabelProps & { class?: HTMLAttributes['class']; inset?: boolean }>()
+const props = defineProps<ContextMenuLabelProps & { class?: HTMLAttributes['class'], inset?: boolean }>()
 
 const delegatedProps = computed(() => {
   const { class: _, ...delegated } = props
@@ -17,7 +17,7 @@ const delegatedProps = computed(() => {
     v-bind="delegatedProps"
     :class="
       cn('px-2 py-1.5 text-sm font-semibold text-foreground',
-         inset && 'pl-8', props.class,
+         inset && 'ps-8', props.class,
       )"
   >
     <slot />
--- a/apps/www/src/lib/registry/default/ui/context-menu/ContextMenuSubTrigger.vue
+++ b/apps/www/src/lib/registry/default/ui/context-menu/ContextMenuSubTrigger.vue
@@ -8,7 +8,7 @@ import {
 import { ChevronRight } from 'lucide-vue-next'
 import { cn } from '@/lib/utils'
 
-const props = defineProps<ContextMenuSubTriggerProps & { class?: HTMLAttributes['class']; inset?: boolean }>()
+const props = defineProps<ContextMenuSubTriggerProps & { class?: HTMLAttributes['class'], inset?: boolean }>()
 
 const delegatedProps = computed(() => {
   const { class: _, ...delegated } = props
@@ -24,11 +24,11 @@ const forwardedProps = useForwardProps(delegatedProps)
     v-bind="forwardedProps"
     :class="cn(
       'flex cursor-default select-none items-center rounded-sm px-2 py-1.5 text-sm outline-none focus:bg-accent focus:text-accent-foreground data-[state=open]:bg-accent data-[state=open]:text-accent-foreground',
-      inset && 'pl-8',
+      inset && 'ps-8',
       props.class,
     )"
   >
     <slot />
-    <ChevronRight class="ml-auto h-4 w-4" />
+    <ChevronRight class="ms-auto h-4 w-4" />
   </ContextMenuSubTrigger>
 </template>
--- a/apps/www/src/lib/registry/default/ui/context-menu/ContextMenuSubTrigger.vue
+++ b/apps/www/src/lib/registry/default/ui/context-menu/ContextMenuSubTrigger.vue
@@ -29,6 +29,6 @@ const forwardedProps = useForwardProps(delegatedProps)
     )"
   >
     <slot />
-    <ChevronRight class="ms-auto h-4 w-4" />
+    <ChevronRight class="ms-auto h-4 w-4 rtl:rotate-180" />
   </ContextMenuSubTrigger>
 </template>
--- a/apps/www/src/lib/registry/default/ui/dropdown-menu/DropdownMenuItem.vue
+++ b/apps/www/src/lib/registry/default/ui/dropdown-menu/DropdownMenuItem.vue
@@ -3,7 +3,7 @@ import { type HTMLAttributes, computed } from 'vue'
 import { DropdownMenuItem, type DropdownMenuItemProps, useForwardProps } from 'radix-vue'
 import { cn } from '@/lib/utils'
 
-const props = defineProps<DropdownMenuItemProps & { class?: HTMLAttributes['class']; inset?: boolean }>()
+const props = defineProps<DropdownMenuItemProps & { class?: HTMLAttributes['class'], inset?: boolean }>()
 
 const delegatedProps = computed(() => {
   const { class: _, ...delegated } = props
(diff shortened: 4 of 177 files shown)


sadeghbarati:
With this, we don't need another CLI feature for converting `ltr` (directional properties) to `rtl` (directional properties)

We can also showcase `ConfigProvider`, `rtl` and `ltr`  in `shadcn-vue` site

CSS Logical Properties support table is nice and most Vite projects are for modern browsers so we are safe to merge this

zernonia:
Im not sure if everyone would like this feature tho. having to change the mindset of `pl` to `ps` is not easy, and might affect the productivity somehow. 

This sounds more like a `opt-in` feature to me. 😅

Saeid-Za:
The alternative would be adding a CLI option with semi-automated for replacing the classes, but it could not be fully automated (some classes have exceptions and have only one directionality). this means the burden of creating multiple version of components would be on maintainers' shoulders.
we already have `default` & `new-york` going forward we'd also need `RTL` versions.


I understand that this creates mental burden for the end user but I had personally forked this repository months ago specially to address multi direction support. (same approach from PR)
What are your thoughts? :eyes:

Saeid-Za:
One other note worth mentioning, this approach is by far the best way to support multi-direction IMO, for reference, here are some alternatives that are just not as good:
- Providing manual multiple versions of the style 
- Using postcss transformer in build step
Quasar uses this [approach](https://github.com/quasarframework/quasar/blob/dev/ui/build/script.build.css.js)  and the plugin is [`postcss-rtl`](https://github.com/vkalinichev/postcss-rtl)

sadeghbarati:
How we can make it an `opt-in` feature?
I think it's better to have multiple registries then

- default directional properties TS
- default directional properties JS

<br />

- default logical properties TS
- default logical properties JS


---

Related issue https://github.com/radix-vue/shadcn-vue/issues/378

lord007tn:
As someone who works with rtl and ltr on daily basis in b2c and b2b products 
i think that logical CSS props are most useful in supporting multi-direction websites better than generating new CSS using plugins like [postcss-rtl](https://github.com/vkalinichev/postcss-rtl)
by default HTML have a LTR direction so we have a default in case your website will be in LTR website
if you are going to support rtl we always tend to change `lang` and `dir` in top `<html>` tag 
and boom we have a rtl direction now 
using libraries that uses css properties that force direction like margin-left, text-left 
will be a headache to manage and that's a no no to use 
as we see css logical props don't cause problems for LTR to use and support RTL also so a win win

zernonia:
Quick question! how does logical properties works with `tailwind-merge`? 🤔

Would it have conflicts if we add `pl-*` to `ps-*`, which one would it resolves?

sadeghbarati:
@zernonia 

[JSPM Playground](https://jspm.org/sandbox#H4sIAAAAAAAAA41UUU/bMBB+51cc3qTCRBIK08RKUlG2ipURhOiqaXvCTdzGxbE922nVIv77Lk6D1k1Ie3Ac++zvvvvuzvF+rjK31gwKV4r+XtxOjOb9PYC4ZI5CVlBjmUtI5WbBGfEGx51g/Yn0cx5HzfrliqQlS8iSs5VWxhHIlHRMIsSK565IcrbkGQv84gi45I5TEdiMCpZ00UEcNQziqcrXHnU/CHACuB7fpXDFJDPUKQOjssaHlGpvHebcweT+pofxOG17UTRvj4YLq8uQq+jN5OR28ePq8vFzemzW4831eNItvg5luvkk7zbD5f1mcuzGjx+76e1oc7xIh4PLXz9Hy8FiMEAfQVDTqRnlfAk8TwjVmvTjCJdbC8Q2M1w7qIVNCPcUS6q9cE+e53bTkt52A7cc5WLFZR6UzMwZWshLDPSFvNRlb/fgxUl4EnbRvXXRtJK5YEG5sCEO4pGf95oRRw2rlj4KCsMxpCqvBINxwUvba+VEsha0EusZFwJmqLMSOTMwNWplmbGw4q5QlQP+x3lbab84YHMY0xk1HLofwtPDrWStKNSuZQbWZMlr8TEblJ5VYGtWF93wLDxtAvzLhNewtIyyVhk+5xKTIZVcl6qydUp2AxZcPoJhIiENhMZ/RXMChWGzV8n8t9hR62cn+Y0rn3lo1XqHGkCLm9awMDOq/KcCzrflBFEE3w13DHXmEhpI7Kgcu5YZdgSYH0SkYJmmWOsIx/FAnSPcJCg1Aeqc4dMKbUqCK1iL0pANvRvsUasEC4WaH+zQOzz3dv/Bag+5xJ768i29gQQefB9kglqbkLdPO/dCt/LzQUfb4H3nCDraBGedw2fSH0mrWYaNy5rOedip0Dhq+h6fAf8e/QYfRDqBpwQAAA==)


`tailwind-merge` will resolve when both classes are of the same kind

- `twMerge('pl-4', 'ps-6')` ---> `pl-4 ps-6` **(directional + logical)**
- `twMerge('pl-4', 'pl-6')` ---> `pl-6` **(directional + directional)**
- `twMerge('ps-4', 'ps-6')` ---> `ps-6` **(logical+ logical)**

Saeid-Za:
It seems that according to unocss playground, the output css for this html:

```html
<div class="ms-10 ml-1"></div>
```

Is in this order:

```css
/* layer: default */
.ml-1{margin-left:0.25rem;}
.ms-10{margin-inline-start:2.5rem;}
```

Which means, logical classes always win by default.
But regarding the tailwind-merge behavior, what is the correct answer in logical-directional situations anyway?

The true answer depends on the current direction of the web page. TM cannot remove `ms-10` from calling `twMerge('ms-10', 'ml-1')` because in RTL direction it changes its behavior.

Well, this is getting complicated 😅

Saeid-Za:
Update:
There is a [response](https://github.com/dcastil/tailwind-merge/issues/388#issuecomment-1979515386) from maintainer of `tailwind-merge`: 

> Hey @lord007tn! 👋
> 
> tailwind-merge doesn't resolve conflicts between normal and logical properties because it doesn't have access to the axis along which logical properties are being applied. This axis is dynamic and can be different for any node in the DOM.
> 
> E.g. `ps-6` (applying the padding-inline-start property) would be in conflict with `pl-4` without any other properties being set. But if a parent node applies the CSS `direction: rtl`, `ps-6` wouldn't be in conflict with `pl-4` anymore because `ps-6` then gets applied to the right side instead. And if additionally `writing-mode: vertical-lr` is applied (to a parent), `ps-6` gets applied to the bottom of the node.

Considering this is the **expected** behavior, this is not an issue of whether user could understand `logical properties`, this PR enforces that **ALL** users must use `logical` instead of `directional` (or normal) properties for their style to work.

I'm fully team `logical` here, but this is an unexpected chain of changes. it seems the only valid solution is to provide an other registry for `Logical` values, so the user understands that they need to work with logical props all the time. nice catch @zernonia.

Anything that I missed here?

Saad5400:
Any updates ... ?

apps/www/src/lib/registry/default/ui/context-menu/ContextMenuItem.vue
```diff
@@ -8,7 +8,7 @@ import {
 } from 'radix-vue'
 import { cn } from '@/lib/utils'
 
-const props = defineProps<ContextMenuItemProps & { class?: HTMLAttributes['class']; inset?: boolean }>()
+const props = defineProps<ContextMenuItemProps & { class?: HTMLAttributes['class'], inset?: boolean }>()
```
Haythamasalama, line 11: No need to modify this line.

apps/www/src/lib/registry/default/ui/context-menu/ContextMenuLabel.vue
```diff
@@ -3,7 +3,7 @@ import { type HTMLAttributes, computed } from 'vue'
 import { ContextMenuLabel, type ContextMenuLabelProps } from 'radix-vue'
 import { cn } from '@/lib/utils'
 
-const props = defineProps<ContextMenuLabelProps & { class?: HTMLAttributes['class']; inset?: boolean }>()
+const props = defineProps<ContextMenuLabelProps & { class?: HTMLAttributes['class'], inset?: boolean }>()
```
Haythamasalama, line 6: No need to modify this line.

apps/www/src/lib/registry/default/ui/context-menu/ContextMenuSubTrigger.vue
```diff
@@ -8,7 +8,7 @@ import {
 import { ChevronRight } from 'lucide-vue-next'
 import { cn } from '@/lib/utils'
 
-const props = defineProps<ContextMenuSubTriggerProps & { class?: HTMLAttributes['class']; inset?: boolean }>()
+const props = defineProps<ContextMenuSubTriggerProps & { class?: HTMLAttributes['class'], inset?: boolean }>()
```
Haythamasalama, line 11: No need to modify this line.

apps/www/src/lib/registry/default/ui/dropdown-menu/DropdownMenuItem.vue
```diff
@@ -3,7 +3,7 @@ import { type HTMLAttributes, computed } from 'vue'
 import { DropdownMenuItem, type DropdownMenuItemProps, useForwardProps } from 'radix-vue'
 import { cn } from '@/lib/utils'
 
-const props = defineProps<DropdownMenuItemProps & { class?: HTMLAttributes['class']; inset?: boolean }>()
+const props = defineProps<DropdownMenuItemProps & { class?: HTMLAttributes['class'], inset?: boolean }>()
```
Haythamasalama, line 6: No need to modify this line.

Haythamasalama (commented):
@zernonia 

It would be great to review this PR — it adds much-needed RTL support to this package, which many other packages already provide.

I’ve reviewed it and noticed a few lines that shouldn’t be changed, particularly those related to styles. Also, some `space-x` classes were replaced with `gap`, which might not be necessary.

zernonia:
Hey @lord007tn, thank you for this PR and for pushing the RTL conversation forward early on — it genuinely helped shape how we thought about direction support. 🙏

Closing this now as it has been superseded:

- The repo has been restructured since (the `apps/www` registry this PR patches no longer exists), so the diff is unrecoverable as-is.
- RTL is now supported first-class through the CLI instead of hardcoding logical classes in the registry source:
  - New projects: `npx shadcn-vue@latest create --rtl` (sets `rtl: true` in `components.json`), and the CLI transforms physical → logical classes (plus `rtl:rotate-180` icon flips) at install time.
  - Existing projects: `npx shadcn-vue@latest migrate rtl`.
  - Direction is provided via reka-ui's [`ConfigProvider`](https://reka-ui.com/docs/utilities/config-provider) with `dir="rtl"`.

This approach also sidesteps the `tailwind-merge` physical-vs-logical conflict issue discussed in this thread, since installed components are transformed consistently rather than mixing conventions in source.

@Saad5400 the above answers your question too — a dedicated RTL docs page is on the way. 📖

Issue #376, opened 2024-03-04
[Feature]: Add missing examples from source


Saeid-Za:
### Describe the feature

There are some missing examples that was recently added to `shadcn/ui`.
I had manually checked the examples and here are the items with new examples:
- [x] Dropdownmenu
- [x] Select
- [x] Skeleton

There is this issue of drifting from new updates when time passes, I'm thinking of an automated way of retrieving the updated file list to manually check and update our source at a later stage, so no updates would be missed.

### Additional information

- [x] I intend to submit a PR for this feature.
- [x] I have already implemented and/or tested this feature.

sadeghbarati:
@Saeid-Za thanks, Can you add the missing Skeleton example from `shadcn-ui`

https://ui.shadcn.com/docs/components/skeleton#card

sadeghbarati:
Only `Select` scrollable remains feel free to do it in your free time

But keep that in mind you can include all docs changes in one PR

Saeid-Za:
Honestly, I had not planned to do them myself 😅

sadeghbarati:
> I intend to submit a PR for this feature.



sorry xD I saw this as checked lol (in my mind) xdd

Saeid-Za:
This should be enough for now, but the main problem remains, how to get notified of updates from shadcn repo.

zernonia:
This would be interesting feature!! Wonder if there's any GitHub Action that do this 🤔

PR #382, opened 2024-03-05
feat: adding skeleton card example

Saeid-Za:
Related issue: #376


sadeghbarati:
It's about documentation so will mark the commit as `docs:`

PR #389, opened 2024-03-07
docs: adding mail example, resolves #373

### LINKED ISSUE #373: [Feature]: Add Mail example to main page
### Describe the feature

We're missing the [mail](https://ui.shadcn.com/examples/mail) component which is blocked by #220. This is a tracking issue, so we could add the missing example when all dependencies are satisfied.

### Additional information

Saeid-Za:
close #373

The one issue here:
- `ScrollArea` has overflowed content in `mail` example, which I think is related to `radix-vue, not sure about this one. the [original example](https://ui.shadcn.com/examples/mail) has this issue too.


Saeid-Za:
It seems that I imported from both registries in some files, I'll fix this when in a short time.

sadeghbarati:
Mobile Nav needs a little change to have the same style as other shadcn ports but I'll do it in another PR to have this nice page sooner in our port

Thanks ❤️‍🔥

Saeid-Za:
Oh, i didn't check the smaller viewports 😅

PR #399, opened 2024-03-08
fix: resizable demo not working

shahzod-davlatov:
[Resizable demo](https://www.shadcn-vue.com/docs/components/resizable) crashed with error **No group found for id "radix-vue-splitter-group-25"**


sadeghbarati:
Thanks ❤️

PR #413, opened 2024-03-15
docs: adding contribution guide

### LINKED ISSUE #409: [Docs]: Adding Detailed Contribution Guide
### Describe the feature

There are several patterns used across this repository, which can make contributing difficult for newcomers without a guide. It's better to have a dedicated and detailed contribution guide to attract more contributors.

### Additional information

Saeid-Za:
Please edit or add any other parts that you see fit.

resolves #409.


apps/www/src/content/docs/contribution.md
```diff
@@ -0,0 +1,221 @@
...
+  </AccordionRoot>
+</template>
+```
+
+As you can see, `AccordionRootEmits` and `AccordionRootProps` types are imported from radix, combined with `useForwardPropsEmits` and then are binded using `v-bind` syntaxt.
+
+### CSS Classes
+There are cases when we want to accept `class` as a prop in our `shadcn/vue` components and then combine it with a default style on our `radix-vue` component.
+
+In these cases, we can not use `v-bind` while `class` is declared as a prop, because this would lead in [double class binding](https://github.com/radix-vue/shadcn-vue/pull/241), so the `class` propert must be dealt specefically.
```
sadeghbarati: > In these cases, we can not use `v-bind` while `class` is declared as a prop, because this would lead in [double class binding](https://github.com/radix-vue/shadcn-vue/pull/241), so the `class` propert must be dealt specefically.

We can use `v-bind` when class is declared as prop 

we only use this way to prevent class duplication with `cn` function

sadeghbarati:
This is so nice Saeid 💯 

Also I would add these changes to [`CONTRIBUTING.md`](https://github.com/radix-vue/shadcn-vue/blob/dev/CONTRIBUTING.md) too

apps/www/src/content/docs/contribution.md
zernonia: Yup. As long as we have `class` as prop, we can use `v-bind` 😁

Saeid-Za: Thanks for the feedback ! I tried to be more clear on this one.

Saeid-Za: I also added some additional information based on `CONTRIBUTING.md`.

Saeid-Za:
Some information in `CONTRIBUTING.md` was wrong, I also updated those.

sadeghbarati (approved):

PR #420, opened 2024-03-21
Fix bug in code example combobox.md

devbyray:
The default value of `const value = ref({})` should contain a falsy value in order to show the "Select framework...".

image

image

When giving it an `{}` empty object will result in a truthy value which is not what we want. When given an a falsy value like `null` it will display the correct text in the button.

### DIFF
--- a/apps/www/src/content/docs/components/combobox.md
+++ b/apps/www/src/content/docs/components/combobox.md
@@ -82,7 +82,7 @@ const value = ref('')
               <Check
                 :class="cn(
                   'mr-2 h-4 w-4',
-                  value?.value === framework.value ? 'opacity-100' : 'opacity-0',
+                  value === framework.value ? 'opacity-100' : 'opacity-0',
                 )"
               />
               {{ framework.label }}
--- a/apps/www/src/content/docs/components/combobox.md
+++ b/apps/www/src/content/docs/components/combobox.md
@@ -82,7 +82,7 @@ const value = ref('')
               <Check
                 :class="cn(
                   'mr-2 h-4 w-4',
-                  value === framework.value ? 'opacity-100' : 'opacity-0',
+                  value?.value === framework.value ? 'opacity-100' : 'opacity-0',
                 )"
               />
               {{ framework.label }}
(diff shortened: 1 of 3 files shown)


apps/www/src/content/docs/components/combobox.md
```diff
@@ -50,7 +50,7 @@ const frameworks = [
 ]
 
 const open = ref(false)
-const value = ref({})
+const value = ref(null)
```
sadeghbarati: I think an empty string will do the trick

```suggestion
const value = ref('')
```

apps/www/src/content/docs/components/combobox.md
```diff
@@ -62,7 +62,7 @@ const value = ref({})
         :aria-expanded="open"
         class="w-[200px] justify-between"
       >
-        {{ value ? frameworks.find((framework) => framework.value === value)?.label : 'Select framework...' }}
+        {{ value ? frameworks.find((framework) => framework.value === value?.value)?.label : 'Select framework...' }}
```
sadeghbarati: based on [`ComboboxDemo.vue`](https://github.com/radix-vue/shadcn-vue/blob/dev/apps/www/src/lib/registry/default/example/ComboboxDemo.vue)

There is no need to use `.value` in `<template>`

```suggestion
        {{ value ? frameworks.find((framework) => framework.value === value)?.label : 'Select framework...' }}
```

sadeghbarati (changes requested):

devbyray:
@sadeghbarati thanks for your suggestions. It appears that the Codesandbox example is different then the one in the Markdown file: https://codesandbox.io/p/devbox/shadcn-vue-comboboxdemo-368j7c?file=%2Fsrc%2FApp.vue

The `:value="framework"` in the Markdown example uses the whole object as value. In the Codesandbox it is `:value="framework.value"`. 

image

apps/www/src/content/docs/components/combobox.md
```diff
@@ -82,7 +82,7 @@ const value = ref({})
               <Check
                 :class="cn(
                   'mr-2 h-4 w-4',
-                  value === framework.value ? 'opacity-100' : 'opacity-0',
+                  value?.value === framework.value ? 'opacity-100' : 'opacity-0',
```
sadeghbarati:
```suggestion
                  value === framework.value ? 'opacity-100' : 'opacity-0',
```

sadeghbarati (approved):

devbyray:
@sadeghbarati I'm to see the fix being merged 💪

PR #426, opened 2024-03-22
Update nuxt.md

kralion:
Added the installation before the initialization of the command `npx shadcn-vue init`


sadeghbarati:
Will check the steps on fresh Nuxt (TS/JS) and will update the installation if it's needed

PR #435, opened 2024-03-25
fix(hover-card): add emits from `radix-vue`

Artemiy111:


sadeghbarati:
@Artemiy111 Thanks, can you run 

```
pnpm run build:registry
``` 

in **apps/www** directory

sadeghbarati (approved):
🙏

PR #442, opened 2024-03-28
fix(example): select language

romanslonov:
Hello, 

Just discovered that the language selector is a little broken now, so this little PR to fix it. 

**Before:**
CleanShot 2024-03-28 at 20 12 26@2x

**After:**
CleanShot 2024-03-28 at 20 14 28@2x
CleanShot 2024-03-28 at 20 14 59@2x


sadeghbarati:
Thanks for the fix since it's related to the docs examples gonna label it as docs

sadeghbarati (approved):

PR #447, opened 2024-04-01
fix(cli): add an option to `component.json` to resolve nuxt buildDir

### LINKED ISSUE #429: [Bug]: `shadcn-nuxt` when the `buildDir` is changed an error is thrown `ENOENT: no such file or directory, stat 'PROJECT_PATH/.nuxt/tsconfig.json'`
### Reproduction

https://stackblitz.com/edit/github-gxjekn?file=nuxt.config.ts

### Describe the bug

when changing the `buildDir` from `nuxt.config.ts` the cli throws an error when generating new components `ENOENT: no such file or directory, stat 'PROJECT_PATH/.nuxt/tsconfig.json'`.

Screenshot 2024-03-23 at 11 24 05

### System Info

```bash
- Operating System: `Darwin`
- Node Version:     `v20.11.0`
- Nuxt Version:     `3.11.1`
- CLI Version:      `3.11.1`
- Nitro Version:    `2.9.4`
- Package Manager:  `npm@10.2.4`
- Builder:          `-`
- User Config:      `devtools`, `app`, `modules`, `shadcn`
- Runtime Modules:  `@nuxtjs/tailwindcss@6.11.4`, `shadcn-nuxt@0.10.2`
- Build Modules:    `-`
```

### Contributes

- [X] I am willing to submit a PR to fix this issue
- [x] I am willing to submit a PR with failing tests

samk-dev:
Adds the option to set `buildDir` for nuxt if it's changed from `nuxt.config`.

It might be missing to be set from the prompt when running `npx shadcn-vue@latest init` if it's necessary please let me know so I can take a look at it asp.

closes #429


zernonia:
Thanks @samk-dev ! Adding `buildDir` would help for your case. However I think let's offer this flexibility to other framework too. Currently the `tsconfig` is expected to be placed in certain folder, I believe adding `tsConfigPath` config option would be a solution. 😁

samk-dev:
Welcome :)

yes adding a `tsConfigPath` makes more sense to support Astro and Laravel as well. I will check the build settings for each framework and make an update the pr. thanks for the input 🙂

samk-dev:
hey @zernonia I am back 🙂

I have checked the rest of the frameworks and saw that `Astro` is almost like `Nuxt` the `buildDir` it can be defined from `astro.config`, so maybe for simplicity we just add an option for `tsConfigPath` instead of 

```ts
 nuxt: z.object({
      buildDir: z.string().default('.nuxt'),
    }).optional()
```

If it's okay with you let me know and will do the changes and update the pr 🙂

thanks

zernonia:
Thanks for looking into that @samk-dev . I agree, let's go with adding `tsConfigPath` as option 😁

samk-dev:
@zernonia I have updated the pr with the needed changes 🙂

- [x] added `tsConfigPath`
- [x] updated tests
- [x] updated docs

PR #448, opened 2024-04-01
chore: adding changelog generator

### LINKED ISSUE #437: [Feature]: Add `changelog` page
### Describe the feature

We're missing the change-log page in website.
Considering other work done on other frameworks, I'd like to implement `changelog` page from [svelte](https://www.shadcn-svelte.com/docs/changelog) version and add previous changelogs to the page.

### Additional information

Saeid-Za:
Adding change log generation via [`changelogen`](https://unjs.io/packages/changelogen).

I've changed the `bumpp` script in `package.json` to use the new generation script.
I've also added the `CHANGELOG.md` file, containing all of the previous releases.

The new `bumpp` command would do the following tasks:
1. Bump the version (Automatic/Manual)
2. Update `CHANGELOG.md` file
3. Generate `changelog` for automatic release PR description.
4. Push the new PR and setting it to `draft` status

from this point onward, a maintainer would check the PR, change the description if required, then merge it with `dev` branch.

To test this script: 
1. Create a Github token  to access project via API
2. Create/update `.env` file, following `.env.example` structure
3. Create a fork of `shadcn-vue` and change content of `changelog.config.ts` to match the new created fork.
  3.1 This step is really important as we would be manipulating  the original repo if the tester has write access to [shadcn-vue](https://github.com/radix-vue/shadcn-vue)
4. Create a dummy commit following conventional commits instructions.
5. Execute `npm run bumpp`
6. Select the desired new version
7. Done

Please feel free to share your thoughts and suggestions.
What is left is `changelog` page for app, which I'll be working on it soon.


scripts/changelog.ts
```diff
@@ -0,0 +1,143 @@
...
+import { versionBump } from 'bumpp'
+import { type SemverBumpType, generateMarkDown, getCurrentGitBranch, loadChangelogConfig } from 'changelogen'
+import { inc } from 'semver'
+import { $fetch } from 'ofetch'
+import { determineBumpType, getContributors, getLatestCommits, isGitClean, loadWorkspace, updateChangelogFile } from './utils'
+import 'dotenv/config'
+
+const remoteName = 'origin'
+
+// Taken from Nuxt Repo (https://github.com/nuxt/nuxt/tree/main/scripts)
```
zernonia, line 12: Appreciate the reference link here! 💚

zernonia:
Thanks @Saeid-Za for the PR! May I ask what's the main difference between using this `scripts/changelog` and https://github.com/changesets/changesets ya? I'm kinda leaning towards `changesets` for the moment as we can utilize it for better version control, and don't need to worry about maintaining our own scripts.

WDYT @sadeghbarati ?

sadeghbarati:
>  main difference between using this scripts/changelog and https://github.com/changesets/changesets ya?

changelogen's changelog is fancier than changesets changelog 😄 I wonder if we could exact same style from changelogen in changesets

Saeid-Za:
Thanks for the feedback !
Well, this is not solely about styling, let me elaborate.
>  I'm kinda leaning towards changesets for the moment as we can utilize it for better version control, and don't need to worry about maintaining our own scripts.
1. I'm not sure if I understand the version control part, why would it be better? we still have the manual (if needed of course) control over what version we're releasing. the script is using `bumpp`  under the hood and supports granular control.
2. Not that maintaining the script is inherently a bad idea, but I understand that it would require further burden if change is frequent. I don't think this script needs changing that much, furthermore, We will always need a script if we would to automate PR request with changelog, see the [`nuxt`](https://github.com/nuxt/nuxt/releases/tag/v3.11.1) release page.

If you think changeset is a better alternative, we could swap `changelogen` with that, because they are really the same tool (with almost the same API) for the same job, but if we were to add automated PRs for releasing the new version with changelog, some sort of script should exist

Is there anything I missed?

Saeid-Za:
Oh, It seems that I was wrong on one point. This repo has a Bot that generates changelog for github. I'll close the PR since this doesn't bring any value.
We still miss a changelog.md file though.

PR #449, opened 2024-04-01
Fix(example): wrong class at blocks example

ovyan:
Hello, found a small issue with the blocks example.

Before the fix:
image

after the fix:
image


own2pwn:
legend

PR #475, opened 2024-04-07
refactor: delete grave accent(`) in template literals

KumJungMin:
### Suggestion
- The name of props type is string.

```js
const props = defineProps<{
  name: string
}>()
```

<br/>

- so, grave accent(`) is unnecessary in template literals :)
- I suggest to delete grave accent(`)


sadeghbarati:
Without the `#` that link will be a route and that route does not exist, The `#` is mostly for scroll to the block section when you click on it

image

PR #479, opened 2024-04-11
Update toast.md

rookie-orange:
When I tried to pop up a Toast based on the document's content, it failed. So I ran it on Stackblitz, but it still failed. After my investigation, I found that I just needed to add the 'Toaster' tag within the 'template' tag.

### DIFF
--- a/apps/www/src/content/docs/components/toast.md
+++ b/apps/www/src/content/docs/components/toast.md
@@ -45,7 +45,7 @@ import { useToast } from '@/components/ui/toast/use-toast'
 <script setup lang="ts">
 import { Button } from '@/components/ui/button'
 import { useToast } from '@/components/ui/toast/use-toast'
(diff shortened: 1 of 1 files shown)


apps/www/src/content/docs/components/toast.md
```diff
@@ -45,11 +45,13 @@ import { useToast } from '@/components/ui/toast/use-toast'
 <script setup lang="ts">
 import { Button } from '@/components/ui/button'
 import { useToast } from '@/components/ui/toast/use-toast'
+import Toaster from "@/components/ui/toast/Toaster.vue"
```
sadeghbarati:
```suggestion
import { Toaster } from "@/components/ui/toast"
```

sadeghbarati (changes requested):

PR #481, opened 2024-04-12
Update drawer.md

rookie-orange:
It sounds like you've identified an important point to add to the documentation about the vaul-drawer-wrapper attribute being necessary for the zoom effect in the Drawer component's background. Clarifying this in the documentation would be helpful for users encountering the same issue.

### DIFF
--- a/apps/www/src/content/docs/components/drawer.md
+++ b/apps/www/src/content/docs/components/drawer.md
@@ -55,6 +55,7 @@ import {
 ```
 ### Scale Background
 If you want the background to have a zoom effect, you need to add the `vaul-drawer-wrapper` attribute to the root component.
+
 ```html
 <div vaul-drawer-wrapper id="app"></div>
 ```
--- a/apps/www/src/content/docs/components/drawer.md
+++ b/apps/www/src/content/docs/components/drawer.md
@@ -53,6 +53,11 @@ import {
   </Drawer>
 </template>
 ```
+### Scale Background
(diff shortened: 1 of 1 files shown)


apps/www/src/content/docs/components/drawer.md
```diff
@@ -53,6 +53,11 @@ import {
   </Drawer>
 </template>
 ```
+### Scale Background
+If you want the background to have a zoom effect, you need to add the `vaul-drawer-wrapper` attribute to the root component.
+```html
+<div vaul-drawer-wrapper id="app"></div>
```
sadeghbarati:
```suggestion
If you want the background to have a zoom effect, you need to add the `vaul-drawer-wrapper` attribute to the root component.

```html
<div vaul-drawer-wrapper id="app"></div>
```

sadeghbarati (changes requested):

apps/www/src/content/docs/components/drawer.md
```diff
@@ -53,6 +53,12 @@ import {
   </Drawer>
 </template>
 ```
+### Scale Background
+If you want the background to have a zoom effect, you need to add the `vaul-drawer-wrapper` attribute to the root component.
```
sadeghbarati:
```suggestion
### Scale Background

If you want the background to have a zoom effect, you need to add the `vaul-drawer-wrapper` attribute to the root component.
```

sadeghbarati (changes requested):

sadeghbarati:
Thanks and sorry for these request changes I just wanted to fixed the spaces here

PR #496, opened 2024-04-15
docs: multiple month range picker

epr3:
Hello,

This PR adds an example to the `DatePicker` component implementing the behaviour seen [here](https://react-tailwindcss-datepicker.vercel.app/demo).

cc @sadeghbarati for the context and example suggestion


sadeghbarati:
Thanks For Your HARD WORK 💯 ❤️

PR #499, opened 2024-04-17
chore: update incorrect link

kn0wn:
Minor link fix


sadeghbarati:
In DatePicker,  we are still using `Calendar` and `RangeCalendar` inside `Popover` primitive 

> The Date Picker is built using a composition of the `<Popover />` and either the `<Calendar />` or `<RangeCalendar />` components.

sadeghbarati:
@kn0wn Ops sorry I'm a copy paster xd

PR #526, opened 2024-05-03
fix: AutoFormFieldObject not utilising base schema

### LINKED ISSUE #525: [Bug]: Autoform fails to render nested enum with optional or default parameter
### Reproduction

https://stackblitz.com/edit/pkm6gj?file=src%2FApp.vue

### Describe the bug

given the zod schema

```js
let schema = z.object({
    count: z.number(),
    nesteObject: z.object({
      framework: z.enum(["vue","react"]).default("vue")
    }),
})
```

The select renders with an empty dropdown. On the other hand if enum is brought to the top level object it works correctly.

### System Info

```bash
-
```

### Contributes

- [X] I am willing to submit a PR to fix this issue

azriel46d:
fixes #525


zernonia:
Thanks for the fix @azriel46d !!! 💚

Issue #271, opened 2024-01-13
Feature(Nuxt): Pass `components.json` options to the module


MuhammadM1998:
### Describe the feature

Hello 👋

I'd like to suggest reading the components options in the nuxt module from `shadcn.components`. Making all shadcn-vue options in one place.

```ts
export default defineNuxtConfig({
  modules: ['@nuxtjs/tailwindcss', 'shadcn-nuxt'],
  shadcn: {
    /**
     * Prefix for all the imported component
     */
    prefix: '',
    /**
     * Directory that the component lives in.
     * @default "./components/ui"
     */
    componentDir: './components/ui',

    // same values as `components.json`
    components: {
    // 
    }
  }
})
```

### Additional information

- [x] I intend to submit a PR for this feature.
- [ ] I have already implemented and/or tested this feature.

kilisei:
I would also like a ``components.json`` path option like 
```ts
configPath: './configs/components.json'
```
 or similar :)

zernonia:
Sorry this will not be implemented, as it increases the complexity in terms of the cli, which could be troublesome of user that wouldn't rely on the module 🙏🏼

Issue #172, opened 2023-11-14
[Bug]: Vee-validate fails when auto importing Form component 


Zackery747:
### Environment

```bash
Developement/Production OS: Windows 10 19043.1110
Node version: 16.0.0
Package manager: pnpm@8.6.0
Radix Vue version: 1.0.0
Shadcn Vue version: 1.0.0
Vue version: 3.0.0
Nuxt version: 3.0.0
Nuxt mode: universal
Nuxt target: server
CSS framework: tailwindcss@3.3.3
Client OS: Windows 10 19043.1110
Browser: Chrome 90.0.4430.212
```

### Link to minimal reproduction

https://stackblitz.com/edit/nuxt-starter-bsx2ja?file=nuxt.config.ts

### Steps to reproduce

You only have to run the project.

### Describe the bug

Cannot destructure property 'componentField' of 'undefined' as it is undefined.

This is caused by Vee-Validate being used in the Form component. It only happens when the Form components are automatically imported. The Form works perfectly if you manually import the Form components.

### Expected behavior

No error is being thrown.

### Conext & Screenshots (if applicable)

```vue
<script setup lang="ts">
import { h } from 'vue';
import { useForm } from 'vee-validate';
import { toTypedSchema } from '@vee-validate/zod';
import * as z from 'zod';

const formSchema = toTypedSchema(
  z.object({
    username: z.string().min(2).max(50),
  })
);

const { handleSubmit } = useForm({
  validationSchema: formSchema,
});

const onSubmit = handleSubmit((values) => {
  toast({
    title: 'You submitted the following values:',
    description: h(
      'pre',
      { class: 'mt-2 w-[340px] rounded-md bg-slate-950 p-4' },
      h('code', { class: 'text-white' }, JSON.stringify(values, null, 2))
    ),
  });
});
</script>

<template>
  <Form class="w-2/3 space-y-6" @submit="onSubmit">
    <FormField v-slot="{ componentField }" name="username">
      <FormItem v-auto-animate>
        <FormLabel>Username</FormLabel>
        <FormControl>
          <Input type="text" placeholder="shadcn" v-bind="componentField" />
        </FormControl>
        <FormDescription> This is your public display name. </FormDescription>
        <FormMessage />
      </FormItem>
    </FormField>
    <Button type="submit"> Submit </Button>
  </Form>
</template>
```

sadeghbarati:
[...]

PR #528, opened 2024-05-04
feat(nuxt-module): auto inject `lib/utils` & refresh the playground

MuhammadM1998:
This PR is an initial step to improve the nuxt module for `shadcn-vue`. Huge thanks to https://github.com/ineshbose for helping me out!

### Module Changes
- **Injected `lib/utils`** by the module into `buildDir`. Now the CLI don't need to inject this file in Nuxt projects, making the file tree more cleaner. An alias for `@/lib/utils` is created to point the injected file for backward compatibility. Also a soft warning is displayed if a `lib/utils` file is found in the root and recommends to remove it as it's redundant.
- Bumped deps (Notably `@nuxt/module-builder` which resolves an harmless but annoying warning)
  
### Playground Changes
- Fixed the issue of tailwindcss classes not applied because `nuxtjs/tailwindcss` was installed but not added to playground's `modules`
- Generated missing `tailwind.css` by the CLI
- Bumped `Button` component by the CLI
- Refreshed the playground style as it was really dull to me and added a dark mode switch (with tailwindcss and radix-vue/icons which are already dependencies)
<table>
 <tr>
    <th>Before</th>
    <th>After</th>
  </tr>

  <tr>
    <td> </td>
    <td></td>
   </tr> 
</table>

### Next Steps
I'll create an issue for those to discuss them first
- Resolve this https://github.com/radix-vue/shadcn-vue/issues/271 by passing the `components.json` options to the module then reading them from the CLI when in a Nuxt project
- Auto install & configure `nuxtjs/tailwindcss` module
- Auto install & configure `nuxtjs/color-mode` module


sadeghbarati:
@MuhammadM1998 Hey thanks for PR

Please check `shadcn-vue` **form** components here

https://github.com/radix-vue/shadcn-vue/blob/dev/apps/www/src/lib/registry/default/ui/form/index.ts#L1

Related issue https://github.com/radix-vue/shadcn-vue/issues/172

packages/module/src/module.ts
```diff
@@ -29,49 +27,31 @@ export default defineNuxtModule<ModuleOptions>({
...
+      logger.warn('[shadcn-nuxt] `lib/utils.ts` is auto generated by the module and can be safely removed.')
+
+    // Components Auto Imports
+    const componentsPath = resolve(COMPONENT_DIR_PATH)
     nuxt.hook('components:dirs', (dirs) => {
       dirs.unshift({
-        path: resolve(COMPONENT_DIR_PATH),
-        extensions: [],
+        path: componentsPath,
+        extensions: ['.vue'], // Scan `.vue` only and skip `.ts` files as we use them inside components file only
```
sadeghbarati: With this, some of the components that exported directly from `index.ts` will be ignored in auto-import process like shadcn-vue form

MuhammadM1998: Didn't test forms 👀 I'll revert it until a better solution comes up and maybe I could try adding tests

packages/module/src/module.ts
```diff
@@ -29,15 +30,36 @@ export default defineNuxtModule<ModuleOptions>({
   async setup({ prefix, componentDir }, nuxt) {
     const COMPONENT_DIR_PATH = componentDir!
     const ROOT_DIR_PATH = nuxt.options.rootDir
+    const UTILS_ALIAS = '@/lib/utils' // Use the same path from the cli for backward compatibility
```
MuhammadM1998, line 33: This won't break if the user changed `aliases.utils` in `components.json` but it would be rather redundant. We can refactor this when `components.json` options is moved to module options

zernonia (approved):
Nicely done @MuhammadM1998 !! Love the implementation and the "Next Steps" plan for this!

PR #531, opened 2024-05-06
chore(nuxt-module): add critical packages to module's dependencies

### LINKED ISSUE #529: Feature(Nuxt): Add `radix-vue` and other main deps as `dependencies`
### Describe the feature

Hey 👋

There are primary dependencies that shadcn-vue relies on and automatically installed by the CLI
- class-variance-authority
- clsx
- radix-vue
- tailwindcss-animate
- tailwind-merge

I think adding them as `dependencies` to the nuxt module makes sense. They would get installed with the module without cluttering the `package.json`. Similar to other nuxt modules which follows the pattern of a 'one-click-setup'.

I've created a branch locally and it works as expected. Happy to make a PR if this is accepted.

### Additional information

MuhammadM1998:
Resolves https://github.com/radix-vue/shadcn-vue/issues/529

This adds the following packages as `dependencies` so that the user don't need to install them.
- class-variance-authority
- clsx
- radix-vue
- tailwind-merge
- tailwindcss-animate

This means a cleaner package.json, I've updated the playground accordingly
```json diff
"dependencies": {
-  "class-variance-authority": "^0.7.0",
-  "clsx": "^2.1.1",
-  "radix-vue": "^1.7.3",
-  "tailwindcss-animate": "^1.0.7"
-  "tailwindcss-merge": "^2.0.3",
  "shadcn-nuxt": "^0.8.0"
}
```

Note that `clsx` is already a dependency in `class-variance-authority` and we're using the same major version `v2` so it can also be safely removed

Tested locally in the playground & a separate project and it works as expected. run `pnpm why radix-vue` and you'll see a similar dependency tree
Screenshot 2024-05-06 111853

### DIFF
--- a/packages/cli/src/commands/init.ts
+++ b/packages/cli/src/commands/init.ts
@@ -310,6 +310,7 @@ export async function runInit(cwd: string, config: Config) {
   // Install dependencies.
   const dependenciesSpinner = ora('Installing dependencies...')?.start()
 
+  // Starting from `shadcn-nuxt` version 0.10.4, Base dependencies are handled by the module so no need to re-add them by the CLI
   const baseDeps = gte(shadcnNuxt?.version || '', '0.10.4') ? [] : PROJECT_DEPENDENCIES.base
   const iconsDep = config.style === 'new-york' ? ['@radix-icons/vue'] : ['lucide-vue-next']
   const deps = baseDeps.concat(iconsDep).filter(Boolean)
--- a/packages/cli/src/commands/init.ts
+++ b/packages/cli/src/commands/init.ts
@@ -9,6 +9,8 @@ import { z } from 'zod'
 import { addDependency, addDevDependency } from 'nypm'
 import { consola } from 'consola'
 import { colors } from 'consola/utils'
+import { gte } from 'semver'
+import { getProjectInfo } from '../utils/get-project-info'
 import * as templates from '../utils/templates'
 import {
   getRegistryBaseColor,
@@ -239,6 +241,15 @@ export async function promptForConfig(
 export async function runInit(cwd: string, config: Config) {
   const spinner = ora('Initializing project...')?.start()
 
+  // Check in in a Nuxt project.
+  const { isNuxt, shadcnNuxt } = await getProjectInfo()
+  if (isNuxt) {
+    consola.log('')
+    shadcnNuxt
+      ? consola.info(`Detected a Nuxt project with 'shadcn-nuxt' v${shadcnNuxt.version}...`)
+      : consola.warn(`Detected a Nuxt project without 'shadcn-nuxt' module. It's recommended to install it.`)
+  }
+
   // Ensure all resolved paths directories exist.
   for (const [key, resolvedPath] of Object.entries(config.resolvedPaths)) {
     // Determine if the path is a file or directory.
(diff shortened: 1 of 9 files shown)


sadeghbarati:
Since it has same changes from this PR https://github.com/radix-vue/shadcn-vue/pull/528 but with +3 new commits please add some text like 

> [!NOTE]
> This PR depends on https://github.com/radix-vue/shadcn-vue/pull/528 to get merged first

MuhammadM1998:
@sadeghbarati Didn't realize it until I made the PR sorry for that😅 I modified it to include only the changes for this PR

sadeghbarati:
@MuhammadM1998, another point regarding the `shadcn-vue` CLI **init** command, when we execute it in Nuxt projects, it still installs those dependencies. Therefore, ensure to update the `cli` package too

MuhammadM1998:
Oh you're right will push changes later today

packages/cli/src/commands/init.ts
```diff
@@ -299,9 +310,10 @@ export async function runInit(cwd: string, config: Config) {
   // Install dependencies.
   const dependenciesSpinner = ora('Installing dependencies...')?.start()
 
-  const deps = PROJECT_DEPENDENCIES.base.concat(
-    config.style === 'new-york' ? ['@radix-icons/vue'] : ['lucide-vue-next'],
-  ).filter(Boolean)
+  // Starting from `shadcn-nuxt` version 0.10.4, Base dependencies are handled by the module so no need to re-add them by the CLI
+  const baseDeps = gte(shadcnNuxt?.version || '', '0.10.4') ? [] : PROJECT_DEPENDENCIES.base
```
MuhammadM1998: The version should be changed to the release that contains this PR

MuhammadM1998:
@sadeghbarati Refactored the CLI to check if `shadcn-nuxt` is installed and checks its version to ensure that it handles the dependencies for backward compat.

sadeghbarati:
Let's wait for Zernonia's opinion cause he is the author of this module 💯

packages/module/playground/nuxt.config.ts
```diff
@@ -1,5 +1,5 @@
 export default defineNuxtConfig({
-  modules: ['@nuxtjs/tailwindcss', '../src/module'],
+  modules: ['@nuxtjs/tailwindcss', 'shadcn-nuxt'],
```
sadeghbarati, line 2:
```suggestion
  modules: ['@nuxtjs/tailwindcss', '../src/module'],
```

I think this might be better instead of linking module inside playground package.json like 
`"shadcn-nuxt": "file:.."`

MuhammadM1998: I've added it to `package.json` so that the deps is installed. We can do the change you suggested and keep the module entry in `package.json` but I think it would be confusing a bit. Also adding the module name in the config mimics the real usage of the module. What do you think?

zernonia:
Thanks for the PR @MuhammadM1998 .. 

One question tho, if user decide to use other version of Radix Vue, how can they install a different version of it?

MuhammadM1998:
@zernonia Thanks! Users can use overrides to use a different version.

```json
"dependencies": {
  "shadcn-nuxt": "^3.0.0", // Has radix-vue v1.7.3 as a dependency
},
"overrides": {
  "radix-vue": "^2.0.0", // This version is what will be installed
}
```

sadeghbarati:
@MuhammadM1998 Let's remove `radix-vue` from `shadcn-nuxt` module deps?

Most of the users are unaware of overrides module features in packageManagers

sadeghbarati:
Also, let's consider checking this Nuxt PR too: https://github.com/nuxt/nuxt/pull/27155

If it gets merged we don't need `@oxc-parser/wasm` anymore I think

MuhammadM1998:
> @MuhammadM1998 Let's remove `radix-vue` from `shadcn-nuxt` module deps?
> 
> Most of the users are unaware of overrides module features in packageManagers

Normally a user won't need to use overrides. A case where a user would need that is that shadcn-nuxt bumped radix-vue to 1.7.8 for example and it caused problems for the user so then he'd use overrides to install radix-vue 1.7.7. That's a slim chance as you see. I'm okay with removing it from dependencies (maybe move it to peerDeps?) but imo its better this way. I'll wait for you decision. Feel free to push changes too!

sadeghbarati:
@MuhammadM1998 Let's merge it for now for further PRs 🙏

PR #534, opened 2024-05-07
chore: add intellisense for Vue components inside markdown files

sadeghbarati:


apps/www/src/vite-env.d.ts
```diff
@@ -1 +0,0 @@
-/// <reference types="vite/client" />
```
sadeghbarati, line 1: deleted `vite-env.d.ts` cause it's already handled inside vitepress

zernonia (approved):
LGTM!

PR #541, opened 2024-05-09
fix: add tailwind and global CSS overwrite warning

### LINKED ISSUE #512: [Feature]: Add a warning when installing the library on a Nuxt project already using Tailwind
### Describe the feature

I have just installed the library on an existing Nuxt 3 project already using tailwind, after running the CLI to configure the components.json, it added all the CSS variables to my tailwind.config.js as expected, however all my previous work on global styling in that file is GONE - fortunately I had a backup, otherwise it can be really frustrating.

Assuming this is the expected behavior of the CLI, in that case you definitely should consider adding a warning to the docs as well as the CLI questions.

### Additional information

sadeghbarati:
Close #512 

image

### DIFF
--- a/packages/cli/src/commands/init.ts
+++ b/packages/cli/src/commands/init.ts
@@ -99,7 +99,7 @@ export async function promptForConfig(
     {
       type: 'toggle',
       name: 'typescript',
-      message: `Would you like to use ${highlight('TypeScript')} (recommended)?`,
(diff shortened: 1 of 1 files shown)


sadeghbarati:
Didn't know I should use `override` or `overwrite`, commit from `shadcn-svelte` helped me

PR #544, opened 2024-05-10
fix: update `PinInput` form 

hrynevychroman:
Hello! Noticed some issues with PinInput, so open a PR to resolve them.

1. Changed `Form` example to use `setFieldValue`, because previously, when using `setValues` that triggered validation on full `Form` and showed errors to fields that are not filled
2. When trying to clear form with `vee-validate` `resetForm` received some errors with `model-value`, added default value for `modelValue` prop inside `PinInput` and everything works as expected. Thanks @Ninja-in-pyjama for helping with this point


sadeghbarati:
Hey @romanhrynevych, thanks for PR


I'll review it. Meanwhile, run build registry command in `apps/www` workspace

sadeghbarati (approved):
I wasn't aware of `setFieldValue` that is nice 🔥

sadeghbarati:
We also have `setValues` in other form demos can you fix them too? @romanhrynevych 

Let's merge this and do it in another PR

hrynevychroman:
> We also have `setValues` in other form demos can you fix them too? @romanhrynevych
> 
> Let's merge this and do it in another PR

Ok, will open new PR soon ❤️

PR #547, opened 2024-05-12
refactor: change `setValues` to `setFieldValue` for better validation

hrynevychroman:
In this PR changed all `setValues` to `setFieldValue` when we need to set only one field of form for better rules checking


sadeghbarati (approved):
❤️

PR #549, opened 2024-05-13
chore: bump vitest to 1.6.0

Dunqing:
I noticed that the project has several different versions of vitest. We should just install vitest in the root directory


zernonia (approved):
Make sense! Thanks for the PR @Dunqing ! 😁

PR #568, opened 2024-05-24
ci: use corepack to automatically handle pnpm version & use GitHub composite actions

MuhammadM1998:
Currently the pnpm version is hardcoded in workflows file and needs to be updated manually like this [`4fc463a` (#531)](https://github.com/radix-vue/shadcn-vue/pull/531/commits/4fc463a9b1892967edcd46ed32f9d6c03c2d726f) otherwise CI fails. 

This PR does the following
- Enables corepack which automatically uses the specified version by `packageManager` in `package.json`.
- Moves the shared setup (Installing node & enabling corepack) into a [GitHub Composite Action](https://docs.github.com/en/actions/creating-actions/creating-a-composite-action) which is basically a reusable workflow that saves us duplication of steps across the workflows.
- Bumps `actions/checkout` to `v4` for consistency as both `v2` and `v3` were used in different workflows.
- Replaces `npx` with `pnpm dlx` in changelogithub script for consistency as we're using pnpm

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -1,0 +1,16 @@
+name: Setup
+description: Installs Node, Enables Corepack and caches pnpm.
+
+runs:
+  using: composite
+
+  steps:
+    - name: Enable corepack
+      run: corepack enable
+      shell: bash
+
+    - name: Setup node & pnpm
+      uses: actions/setup-node@v4
+      with:
+        node-version: lts/*
+        cache: pnpm
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -48,7 +48,7 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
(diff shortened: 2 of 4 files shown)


.github/actions/setup/action.yml
```diff
@@ -0,0 +1,16 @@
+name: Setup
+description: Installs Node, Enables Corepack and caches pnpm.
+
+runs:
+  using: composite
+
+  steps:
+    - name: Enable corepack
+      run: corepack enable
+      shell: bash
```
sadeghbarati, line 10: Shouldn't this step be after node lts setup? Or it doesn't matter?

MuhammadM1998: Its a workaround related to setup-node action (https://github.com/actions/setup-node/issues/480). Nuxt uses the same setup for reference https://github.com/nuxt/nuxt/blob/main/.github%2Fworkflows%2Fci.yml#L39

sadeghbarati:
@MuhammadM1998 🙌

PR #581, opened 2024-05-29
chore: update linting rules and adding tailwind linter plugin

Saeid-Za:
### ❓ Type of change
- [X] 🧹 Chore (updates to the build process or auxiliary tools and libraries)

### 📚 Description

Current `dev` branch has some incorrect configuration in `eslint.config.js` file.
1. Order of file-specific  & general rules were incorrect, this caused false-positive warnings in linting stage.

I've added `tailwindcss` linting plugins to better align the project with it's styling best practices.

The amount of changes are unfortunately large, but they are the result of running the lint with new configuration.
The main files that should be reviewed are `eslint.config.js` and `package.json`, the rest could be ignored as they could be reproduced by running `eslint` command with this new configuration.

### DIFF
--- a/apps/www/.vitepress/theme/layout/MainLayout.vue
+++ b/apps/www/.vitepress/theme/layout/MainLayout.vue
@@ -209,7 +209,7 @@ watch(() => $route.path, (n) => {
             <span class="inline-block ml-2">
               Ported to Vue by
               <a
-                href="https://github.com/radix-vue"
+                href="https://github.com/unovue"
                 target="_blank"
                 class="underline underline-offset-4 font-bold decoration-foreground"
               >


sadeghbarati:
Hey @Saeid-Za, I appreciate your efforts 🙌


I think eslint-plugin-tailwind already supports flat config (eslint v9)

eg. https://github.com/blefnk/relivator-nextjs-starter/blob/main/eslint.config.ts#L75

sadeghbarati:
Let's ignore eslint `size-*` transformation for `registry/[default,new-york]/ui`?


@zernonia @Saeid-Za wdyt

Saeid-Za:
Thanks for the comment !
I'm in favor of using `size` rule for the repo.
Pros:
- fewer classes lead to ease of understanding
-  reduction in final bundle size by a few bytes.

Cons: 
- Huge amount of changes in the repo, that would make a pain point for updating users
  - since there are no updating strategy, this wouldn't matter right now (?)

PR #603, opened 2024-06-12
fix: change label/value order in tasks priorities data

pereira0x:
### ❓ Type of change

### 📚 Description

This PR addresses and resolves an inconsistency in the data of the task example, ensuring that priority attributes follow the same order as labels and statuses attributes.

### 📝 Checklist

### DIFF
--- a/apps/www/src/examples/tasks/data/data.ts
+++ b/apps/www/src/examples/tasks/data/data.ts
@@ -53,18 +53,18 @@ export const statuses = [
 
 export const priorities = [
   {
-    label: 'Low',
     value: 'low',
+    label: 'Low',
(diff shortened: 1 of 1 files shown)


sadeghbarati:
Hi, can you explain more or record a video of this inconsistency?

pereira0x:
Hey, here you go:
In both `labels` and `statuses`, the order of their attributes is `value` first, then `label`.
image
image

In `priorities`, the `label` comes first, then it's the `value:`
image

I believe it would be nice to use the same order in all of them.

sadeghbarati:
I get it now, but it's not that important right? we should add eslint rule to sort them in alphabetical order

PR #616, opened 2024-06-17
chore: updating source/docs to match the changes in `shadcn/ui` version

Saeid-Za:
### ❓ Type of change

### 📚 Description

I have reviewed the commits in the `shadcn/ui` repository dating back 7 months and have applied any missing changes from that time period.
I'm sure that any commit older than 7 months is already applied here.
There were some missing functionalities that could be investigated.
1. Lift mode
2. Copy Code Button in `ComponentPreview`
3. Some Additional CLI features (like prefixing tailwind class and more)

This PR tries to carve the path for #376 (Previously was: `Automated way of retrieving the updated files from main repo`).
With this PR accepted, we would be in a similar starting point going forward.

To track the changes going forward, we could store the latest commit ID from `shadcn/ui` that `shacn-vue` is synced with, in the repo. (With this PR, the commit ID is: [`06cc0cdf3d080555d26abbe6639f2d7f6341ec73`](https://github.com/shadcn-ui/ui/commit/06cc0cdf3d080555d26abbe6639f2d7f6341ec73)

From [zernonia](https://github.com/zernonia).
> This would be interesting feature!! Wonder if there's any GitHub Action that do this 🤔

Unfortunately, the Github actions wouldn't allow execution of code on [trigger events](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows) of an other Repository that we don't have access to, so we're back to manually checking if the `main` branch of `shadcn/ui` is committed on, which could be automated by a script (and also leveraging the commit ID discussed before) but the execution of this script is manual.

One potential solution could be to set up a cron job as a GitHub action for this repository in order to execute the script automatically.

Please share your opinion and let me know if you think this topic should be addressed in a separate issue.

### 📝 Checklist


sadeghbarati (approved):

zernonia (approved):
LGTM!! Thanks for tracing all the changes in `shadcn/ui` and populate it here @Saeid-Za ! 😁

PR #636, opened 2024-06-28
fix: add chartRef to solve gradiant in same color issue

arvidsson-geins:
### ❓ Type of change

### 📝 Checklist


apps/www/src/lib/registry/default/ui/chart-area/AreaChart.vue
```diff
@@ -41,6 +41,8 @@ const emits = defineEmits<{
 type KeyOfT = Extract<keyof T, string>
 type Data = typeof props.data[number]
 
+const chartRef = Math.random().toString(36).substring(2, 9)
```
sadeghbarati: `useId` might be a better way

arvidsson-geins: True! Updated to useId() to generate chartRef

sadeghbarati:
Thanks

PR #648, opened 2024-07-07
docs: add sub components example

Lol1chw:
### ❓ Type of change

### 📚 Description

Improves dx. Provides an example of how to make a component reusable.
Made changes to the passed component types to not be tied to a specific data model, now the type is based on a generic type.
Extended `tasks.json` for use in the example, the data for `tasks.json` is taken from DummyJSON.
Added a user to the schema for tasks.json.
Fixed filter-function type incompatibility in `examples/tasks/components/DataTableFacetedFilter.vue`

### 📝 Checklist

### DIFF
--- a/apps/www/src/content/docs/components/data-table.md
+++ b/apps/www/src/content/docs/components/data-table.md
@@ -866,11 +866,25 @@ This adds a checkbox to each row and a checkbox in the header to select all rows
 
 You can show the number of selected rows using the `table.getFilteredSelectedRowModel()` API.
 
-```vue
-<div class="flex-1 text-sm text-muted-foreground">
-    {{ table.getFilteredSelectedRowModel().rows.length }} of
-    {{ table.getFilteredRowModel().rows.length }} row(s) selected.
-</div>
+```vue:line-numbers {8-11}
+<template>
+  <div>
+    <div class="border rounded-md">
+        <Table />
+    </div>
+
+    <div class="flex items-center justify-end space-x-2 py-4">
+      <div class="flex-1 text-sm text-muted-foreground">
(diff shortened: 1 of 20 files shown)


sadeghbarati:
Can you share an image or video about the PR changes, thanks

Lol1chw:
Video:
https://github.com/radix-vue/shadcn-vue/assets/94941930/859f1e46-e080-4fd4-8f75-8084cf7f25bf

Lol1chw:
Btw, I found the real reason related to the type incompatibility  in the filter-function of the `Command.vue` component. The generic type was not passed to the ComboboxRoot<T> type, so filter-function did not know what type it should expect.

So, if you pass in `Command.vue` a generic that inherits from the AcceptableValue type
` type AcceptableValue = string | number | boolean | Record<string, any>;`

```vue
<script setup lang="ts" generic="T extends string | number | boolean | Record<string, any>">
const props = withDefaults(defineProps<ComboboxRootProps<T> & { class?: HTMLAttributes['class'] }>(), {
  open: true,
})
```


then filter-function will automatically output the type for 'list':

image

This way we can get rid of the type predicate, which is always not completely type-safety.
We'll also reduce the amount of code.

sadeghbarati:
@Lol1chw Thanks can you do the same thing to `Command.vue` in the default style and after that run `pnpm build:registry` in `apps/www`

By the way, types will be better in v2, and thanks for the fix

Lol1chw:
My bad, I forgot about the `default` variant. Made changes to the default. 
Build was successful without errors.
image

sadeghbarati:
I think showcasing `subRow` on Data Table page would be enough, thanks

Issue #663, opened 2024-07-17
[Bug]: Donut chart index prop does not render if value is different than "name"


decito:
### Reproduction

https://codesandbox.io/p/devbox/shadcn-vue-donutchartdemo-forked-8z4mt4

### Describe the bug

Basically, unless the `index` property is equal to "name", it does not render the text passed in the property.

Inspecting the code closely, the bug is due to the fact that the `const omittedData` of the **ChartSingleTooltip** component is assigning the `name` value to `data.name` and not for the value passed in the `index` prop of the **DonutChart** component.

I managed to fix the bug in my project by changing the expected to `Object.values(data)[0]`. I don't have enough knowledge to understand whether this would be the best course of action, nor have I checked whether the ChartSingleTooltip component is being used by another and whether this change breaks the behavior in other imports. Nonetheless, I'll submit a PR.

In the Reproduction, I described two objects, one with the name property and the other with any other. Hovering the chart, one of the names are rendered while the other one isn't.

This is my first bug report here on GitHub :)

### System Info

```bash
System:
    OS: Linux 6.1 Debian GNU/Linux 12 (bookworm) 12 (bookworm)
    CPU: (2) x64 AMD EPYC
    Memory: 1.78 GB / 4.01 GB
    Container: Yes
    Shell: Unknown
Binaries:
    Node: 18.20.4 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 10.7.0 - /usr/local/bin/npm
npmPackages:
    @vueuse/core: ^10.11.0 => 10.11.0 
    radix-vue: ^1.9.0 => 1.9.1 
    shadcn-vue: latest => 0.10.5 
    vue: latest => 3.4.32
```

### Contributes

- [X] I am willing to submit a PR to fix this issue
- [X] I am willing to submit a PR with failing tests

fds44866:
Hey, could you please double-check the access rights of the `codesandbox`?

```
Devbox not found
It's likely that the Devbox you're trying to access doesn't exist or you don't have the required permissions to access it.
If the issue persists, please contact our support team at [support@codesandbox.io](mailto:support@codesandbox.io) for further assistance, making sure to copy the error details below.


Problem details and configurations
```

entl:
Hey, could you please double-check access rights to the `codesandbox`, thanks.

```
Devbox not found
It's likely that the Devbox you're trying to access doesn't exist or you don't have the required permissions to access it.
[...]

PR #664, opened 2024-07-17
fix: fix name rendering in ChartSingleTooltip component

decito:
### 🔗 Linked issue

#663

### ❓ Type of change

### 📚 Description

I managed to fix the bug described in #663 by changing the expected to Object.values(data)[0]. I don't have enough knowledge to understand whether this would be the best course of action, nor have I checked whether the ChartSingleTooltip component is being used by another and whether this change breaks the behavior in other imports. Nonetheless, I'll submit the PR.

### 📝 Checklist


decito:
Few Updates:

- Apply the same fix in the "New York" component.
- Checked the usage of the component in question, and it doesn't appear to be used anywhere else.
- Also noticed the presence of the styles files and updated those as well.

decito:
I just noticed that this solution in unoptimal because, in the fix proposed, the name to render would always need to be the first property in the object passed in the `data` prop.  I'll keep looking for a better way to always display the value passed in the index prop

apps/www/src/lib/registry/new-york/ui/chart/ChartSingleTooltip.vue
```diff
@@ -43,7 +43,7 @@ function template(d: any, i: number, elements: (HTMLElement | SVGElement)[]) {
     }
     else {
       const style = getComputedStyle(elements[i])
-      const omittedData = [{ name: data.name, value: props.valueFormatter(data[props.index]), color: style.fill }]
+      const omittedData = [{ name: Object.values(data)[0], value: props.valueFormatter(data[props.index]), color: style.fill }]
```
zernonia, line 46: The `items` props is expecting type `BulletLegendItemInterface[]`, which has name as the primary identifier. 
I can't load your codesandbox and not sure what issue you are running into.


```
interface BulletLegendItemInterface {
    name: string | number;
    color?: string;
    className?: string;
    shape?: BulletShape;
    inactive?: boolean;
    hidden?: boolean;
    pointer?: boolean;
}
```



**Cant load this**

https://codesandbox.io/p/devbox/shadcn-vue-donutchartdemo-forked-8z4mt4

decito: Hi, I've granted permission to the codesandbox project :)

PR #677, opened 2024-07-24
feat(Dialog): add prop `teleported` to disable teleport

ntnyq:
### ❓ Type of change

### 📚 Description

This PR adds a prop `teleported` to Dialog so that the default teleport behavior can be disabled by setting it to `false`.

Problems may resolved:

1. Opening a dialog in a ShadowRoot which doesn't share styles with main context. By default, the dialog will be teleported to main context.
2. After using requestFullscreen API with an element but not `body` . Then opening the dialog.

### 📝 Checklist


sadeghbarati:
It should be done in the userland, if you don't want to use the portal create another UI path like

```
components/ui/dialog (add DialogPortal component)
components/ui/dialog-without-portal (don't add DialogPortal component)
```

zernonia:
Agree with @sadeghbarati . This prop should be handled by the user itself, and not `shadcn-vue`. As this library allows user to customize the behavior to their use case, adding this prop is unnecessary and not everyone need this. 

Still thanks for your PR @ntnyq !

Issue #716, opened 2024-08-17
[Bug]: event type of autoform submit is strange


sinh117801:
### Reproduction

https://www.shadcn-vue.com/docs/components/auto-form.html

### Describe the bug

It is somewhat strange to me when using AutoForm.

```vue
<template>
  <AutoForm
    :class="cn('grid gap-6', $attrs.class ?? '')"
    :schema="schema"
    :field-config="{
      password: {
        label: 'Your secure password',
        inputProps: {
          type: 'password',
          placeholder: '••••••••',
        },
      },
    }"
    @submit="handleSubmit"
  >
    <Button type="submit">
      Submit
    </Button>
  </AutoForm>
</template>

<script setup lang="ts">
import z from 'zod';
import { h } from 'vue';
import { cn } from '@/utils';
import { Button } from '@/components/shares/ui/button';
import { toast } from '@/components/shares/ui/toast';
import { AutoForm } from '@/components/shares/ui/auto-form';

const schema = z.object({
  username: z
    .string({
      required_error: 'Username is required.',
    })
    .min(2, {
      message: 'Username must be at least 2 characters.',
    }),

  password: z
    .string({
      required_error: 'Password is required.',
    })
    .min(8, {
      message: 'Password must be at least 8 characters.',
    }),
});

function handleSubmit(values: Record<string, any>) {
  toast({
    title: 'You submitted the following values:',
    description: h('pre', { class: 'mt-2 w-[340px] rounded-md bg-slate-950 p-4' }, h('code', { class: 'text-white' }, JSON.stringify(values, null, 2))),
  });
}
</script>
```

The snippet code above just follows the basic example in the docs. The strange thing is that when I assign the type to `values: z.infer<typeof schema>` in the function `handleSubmit` instead of `values: Record<string, any>`, I get an error.

![image](https://github.com/user-attachments/assets/3139eff4-298a-4582-b1eb-8d4d9eb70367)

I think it would be better to allow the case to assign `values: z.infer<typeof schema>`, so that in the function `handleSubmit`, I can still utilize the type of values. I have fixed it on my side and I am willing to contribute that change.

### System Info

```bash
System:
  OS: Linux 5.15 Ubuntu 22.04.4 LTS 22.04.4 LTS (Jammy Jellyfish)
  CPU: (20) x64 12th Gen Intel(R) Core(TM) i7-12700H
  Memory: 3.01 GB / 7.62 GB
  Container: Yes
[...]

PR #717, opened 2024-08-17
enhancement: update emits submit event type of autoform (#716)

sinh117801:
If compiler options are only added to just tsconfig.json, module resolution fails for `@/lib/utils` during build.

### ❓ Type of change

### 📝 Checklist


zernonia:
Since both of them yield the same types for `@submit`, we can apply this change 😁

PR #723, opened 2024-08-20
doc: Update Tags with Combobox example to make conent collapsable

NehemiahAklil:
### ❓ Type of change

### 📚 Description

Tags with Combobox example on the documentation isn't in accordance to the [Combobox documentation of radix-vue](https://www.radix-vue.com/components/combobox) which wraps the Combobox content inside ComboboxContent before wrapping in ComboboxPortal component which makes it collapsible as the demo below. check the current docs and you will see once it pops down the list of options there is no option to dismiss the options. Here is the anatomy of Combobox the current commit follows as the radix-vue documentation instructs.
```javascript
<template>
  <ComboboxRoot>
    <ComboboxAnchor>
      <ComboboxInput />
      <ComboboxTrigger />
      <ComboboxCancel />
    </ComboboxAnchor>

    <ComboboxPortal>
      <ComboboxContent>
        <ComboboxViewport>
          <ComboboxEmpty />

          <ComboboxItem>
            <ComboboxItemIndicator />
          </ComboboxItem>

          <ComboboxGroup>
            <ComboboxLabel />
            <ComboboxItem>
              <ComboboxItemIndicator />
            </ComboboxItem>
          </ComboboxGroup>
          <ComboboxSeparator />
        </ComboboxViewport>

        <ComboboxArrow />
      </ComboboxContent>
    </ComboboxPortal>
  </ComboboxRoot>
</template>
```
### 📝 Checklist


sadeghbarati:
Thanks, by the way, you can upload files without adding them to your repository

NehemiahAklil:
Sorry about that I was in hurry. And I didn't think it would add that last commit to the PR only the one commit with the fix.

xak2000:
Can anybody explain please why does this fix work?

I mean, I tested it and it really works! Wrapping `CommandList` in `ComboboxContent` helps to make the popover closable by pressing <kbd>Esc</kbd> and clicking outside. Without this wrapping these actions do nothing.

But as I see the `CommandList` already includes `ComboboxContent` component.

Or, rather I would say, [`CommandList` is `ComboboxContent`](https://github.com/unovue/shadcn-vue/blob/b7ef4653f75ea30b28ec2ff635cda4ab71dc2984/apps/www/src/lib/registry/default/ui/command/CommandList.vue#L22-L26)!

So, by wrapping `CommandList` in `ComboboxContent` we basically create a structure like this:

```
<ComboboxContent>
  <ComboboxContent> <!-- This one is from CommandList template -->
  </ComboboxContent>
</ComboboxContent>
```

Even `ComboboxContent` properties such as `avoid-collisions` work if added to `CommandList`. They do not work if I add them to the wrapping `ComboboxContent`, btw.

So, I'm confused.. 😕 

Also according to the anatomy of Combobox component above, `ComboboxViewport` should be present too, but it is missing, nevertheless it doesn't make the example broken for some reason. Adding/removing `ComboboxViewport` doesn't change anything at all. 🤔

PR #844, opened 2024-10-29
docs: Accordion. Update keyframes config to be compatible with latest tailwind config …

rizhenkov:
…format

### ❓ Type of change

### 📚 Description

Old format [described in docs](https://www.shadcn-vue.com/docs/components/accordion) (from - to) makes errors in TS. 
Ref: https://tailwindcss.com/docs/animation#customizing-your-theme

before:
image
after:
image

### 📝 Checklist

### DIFF
--- a/apps/www/src/content/docs/components/accordion.md
+++ b/apps/www/src/content/docs/components/accordion.md
@@ -28,12 +28,12 @@ module.exports = {
     extend: {
       keyframes: {
         'accordion-down': {
-          from: { height: 0 },
-          to: { height: 'var(--radix-accordion-content-height)' },
+          '0%': { height: '0' },
+          '100%': { height: 'var(--radix-accordion-content-height)' },
         },
         'accordion-up': {
-          from: { height: 'var(--radix-accordion-content-height)' },
-          to: { height: 0 },
+          '0%': { height: 'var(--radix-accordion-content-height)' },
+          '100%': { height: '0' },
         },
       },
       animation: {


xak2000:
There is nothing wrong with `from` and `to`.

`keyframes` property is defined like this:

```
keyframes?: ResolvableTo<KeyValuePair<string, KeyValuePair<string, KeyValuePair<string, string>>>>
```

The only problem with the current docs is that the last part is `KeyValuePair<string, string>`, so `height` cannot be a number.

`{ height: 0 }` should be replaced with `{ height: '0' }`. That's all.

I don't see a reason of changing `from` and `to` to `0%` and `100%`.

xak2000:
Also, I think, maybe it's worth to fix this problem not just for `Accordion` docs, but for every component's docs ([that has `keyframes` in the code](https://github.com/search?q=repo%3Aunovue%2Fshadcn-vue%20keyframes&type=code))?

Most importantly for https://github.com/unovue/shadcn-vue/blob/75a5bce92f0a5ac035c63b27f7de4dcab30d5cb2/packages/cli/src/utils/templates.ts#L127 that is used as a template for `tailwind.config.js` generation when `npx shadcn-vue@latest init` is executed.

zernonia:
Thanks for the PR, however we are following the same code from [ui.shadcn](https://ui.shadcn.com/docs/components/accordion) 😁

xak2000:
@zernonia So, what is the solution? Will you create a new PR or issue to track this problem? I mean, even if this PR is not correct, the problem is still real. 😄 

The page you linked contains examples (see "Manual installation"):

```
...
from: { height: "0" }
...
```

So, the `hieght` is `string`, not `number`.

Therefore, following the same code from [ui.shadcn](https://ui.shadcn.com/docs/components/accordion) we should also use a string here. :)

And not only for Accordion, but in any other components/templates.

PR #906, opened 2024-11-16
Refactor: Utilize Vue's built-in defineModel for model implementation

PiotrWasak:
### ❓ Type of change

### 📚 Description

This PR refactors the code to use Vue's built-in defineModel instead of relying on custom libraries. By leveraging Vue's native functionality:

The codebase is simplified with less code.
External library dependencies are no longer needed for this functionality.

### 📝 Checklist

### DIFF
--- a/apps/www/src/lib/registry/default/ui/textarea/Textarea.vue
+++ b/apps/www/src/lib/registry/default/ui/textarea/Textarea.vue
@@ -1,22 +1,14 @@
 <script setup lang="ts">
 import type { HTMLAttributes } from 'vue'
 import { cn } from '@/lib/utils'
-import { useVModel } from '@vueuse/core'
 
 const props = defineProps<{
   class?: HTMLAttributes['class']
   defaultValue?: string | number
-  modelValue?: string | number
 }>()
 
-const emits = defineEmits<{
-  (e: 'update:modelValue', payload: string | number): void
-}>()
-
-const modelValue = useVModel(props, 'modelValue', emits, {
-  passive: true,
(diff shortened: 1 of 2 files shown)


apps/www/src/lib/registry/default/ui/textarea/Textarea.vue
```diff
@@ -1,22 +1,14 @@
...
-const emits = defineEmits<{
-  (e: 'update:modelValue', payload: string | number): void
-}>()
-
-const modelValue = useVModel(props, 'modelValue', emits, {
-  passive: true,
-  defaultValue: props.defaultValue,
-})
+const modelValue = defineModel<string | number>()
+modelValue.value = props.defaultValue
```
sadeghbarati, line 11:
```suggestion
const modelValue = defineModel<string | number>({
  default: props.defaultValue
})
```

sadeghbarati:
I think it is soon to include this macro in the components cause this macro added from 3.4+

Let's see what other's opinion is

apps/www/src/lib/registry/default/ui/textarea/Textarea.vue
PiotrWasak: It cannot be done like that, already tried it :grin: 
"[@vue/compiler-sfc] `defineModel()` in <script setup> cannot reference locally declared variables because it will be hoisted outside of the setup() function."

zernonia:
I think due to Vue's current limitation (defineModel can't access props.defaultValue), we can't move forward with this PR. I'll be closing it 🙏🏼

PR #965, opened 2024-12-17
Allow to pass label and description via props

thewebartisan7:
### 🔗 Linked issue

[Issue](https://github.com/unovue/shadcn-vue/issues/964)

### ❓ Type of change

- [ ] 📖 Documentation (coming)
- [X] 👌 Enhancement (improving an existing functionality like performance)

### 📚 Description

Describe the feature
Right now we have many components for input, example:

```vue
          <FormField v-slot="{ componentField }" name="name">
            <FormItem>
              <FormLabel>Name</FormLabel>

              <FormControl>
                <Input
                  type="text"
                  placeholder="Your name"
                  v-bind="componentField"
                />
              </FormControl>

              <FormDescription>
               This is your public display name.
              </FormDescription>

              <FormMessage />
            </FormItem>
          </FormField>
```

We can reduce by passing, optionally, the label and description via props, and also the error is display conditionally instead of description, such avoiding to have description and error which doesn't looks good.

```vue
          <FormField v-slot="{ componentField }" name="name">
            <FormItem>
              <FormControl
                label="Name"
                description="This is your public display name."
               >
                <Input
                  type="text"
                  placeholder="How we can call you?"
                  v-bind="componentField"
                  autofocus
                />
              </FormControl>
            </FormItem>
          </FormField>
```

- [X] I have linked an issue or discussion.

NOTE: I will update docs showing an example if this PR is ok to be merged.

### DIFF
--- a/apps/www/src/lib/registry/default/ui/form/FormControl.vue
+++ b/apps/www/src/lib/registry/default/ui/form/FormControl.vue
@@ -1,6 +1,6 @@
 <script lang="ts" setup>
-import { Slot } from "radix-vue";
-import { useFormField } from "./useFormField";
+import { Slot } from 'radix-vue';
+import { useFormField } from './useFormField';
 
 const props = defineProps<{
   label?: string;
--- a/apps/www/src/lib/registry/default/ui/form/FormControl.vue
+++ b/apps/www/src/lib/registry/default/ui/form/FormControl.vue
@@ -1,16 +1,35 @@
 <script lang="ts" setup>
-import { Slot } from 'radix-vue'
-import { useFormField } from './useFormField'
+import { Slot } from "radix-vue";
+import { useFormField } from "./useFormField";
 
-const { error, formItemId, formDescriptionId, formMessageId } = useFormField()
+const props = defineProps<{
+  label?: string;
+  description?: string;
+}>();
+
+const { error, formItemId, formDescriptionId, formMessageId } = useFormField();
 </script>
 
 <template>
+  <FormLabel v-if="props.label">{{ props.label }}</FormLabel>
+
   <Slot
     :id="formItemId"
-    :aria-describedby="!error ? `${formDescriptionId}` : `${formDescriptionId} ${formMessageId}`"
+    :aria-describedby="
+      !error
+        ? `${formDescriptionId}`
+        : props.description
+        ? `${formMessageId}`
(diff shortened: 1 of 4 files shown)


thewebartisan7:
Result without error:

Screenshot 2024-12-17 at 11 54 36

With error:

Screenshot 2024-12-17 at 11 54 42

thewebartisan7:
I found a problem in my PR. We can't use `const { error } = useFormField();` in FormItem.vue to check if there is an error because the `FORM_ITEM_INJECTION_KEY` is not yet provided at that point.

If we want to hide description when there is an error, then we have to do this inside  `FormDescription.vue` , I don't see other way. For example:


FormDescription

```vue
<script lang="ts" setup>
import { cn } from "@/lib/utils";
import type { HTMLAttributes } from "vue";
import { useFormField } from "./useFormField";

const props = defineProps<{
  class?: HTMLAttributes["class"];
}>();

const { formDescriptionId, error } = useFormField();
</script>

<template>
  <p
    :id="formDescriptionId"
    :class="cn('text-sm text-muted-foreground', props.class)"
    v-if="!error"
  >
    <slot />
  </p>

  <FormMessage v-if="error" />
</template>
```

But this will be breaking changes for form that has already loaded `FormMessage.vue` manually, so we could add a new prop in `FormDescription` like `withError` or `toggleError` which by default is false. Or a new component like `FormItemDescription.vue` (so it's clear that is being loade inside `FormItem.vue`).

But of course this additional feature to toggle error and description is something else, and in case you don't want, we can just keep the other one.

Let me know what do you think.

thewebartisan7:
I found a better way to handle this by using `FormControl.vue `
Looking for to hear from you.

sadeghbarati:
Thanks for the PR and the insight

Let's check the diff between (`props`) and (`components per line` or  `low-level components`)


- low-level components
  - give you control over how you want to wrap them. You're free to introduce your own high-level API to better suit the needs of your team and product. (radix-ui)
  - with low-level components, you have more control to components view, for example, this is a normal label and input component, but if you want to show the label at the bottom of the input, simply move the label to the bottom
  
  ```diff
  - <Label>label top of the input</Label>
  <Input />
  + <Label>label top of the input</Label>
  ```



- Props are good too but you might not have full control of the components view in this way

PR #1028, opened 2025-02-03
docs: Change href to link to the example instead of a non existing page

xibn:
### ❓ Type of change

### 📚 Description

Change href to link to the example instead of a non existing page. This is similar to how the "Login" button in the authentication example works.

### 📝 Checklist


zernonia:
As this is just an example, and for the ease of copy/paste we will stick back to the current link even if it's broken  😁

Thanks for the PR nonetheless

Issue #1208, opened 2025-04-24
[Bug]: Theme preference toggle button's icon disappears on shadcn-vue homepage in production


gungdekrisna:
### Reproduction

Its happen on the shadcn-vue main page: https://www.shadcn-vue.com , steps to reproduce explained in the description below.

## Description

When visiting the [shadcn-vue main page](https://www.shadcn-vue.com/) with **dark mode enabled by default**, switching to light mode using the theme toggle button causes the button's icon to **disappear**.

As a result, the button becomes **visually invisible**, since its icon no longer appears and its background blends with the page.

<img width="1462" alt="Image" src="https://github.com/user-attachments/assets/2618a42d-b1e6-41d3-b510-76ae3fdaa64b" />

&nbsp;
This issue is accompanied by the following errors in the **Chrome DevTools** console:

<img width="978" alt="Image" src="https://github.com/user-attachments/assets/70deb816-020e-49e7-ba78-8e275bf30991" />

---

## Steps to Reproduce

1. Open [https://www.shadcn-vue.com/](https://www.shadcn-vue.com/) in a browser **with dark mode enabled**.
   - (If you're in light mode, switch to dark mode first, then reload the page.)
2. Click the **theme toggle** button to switch to light mode.

<img width="1296" alt="Image" src="https://github.com/user-attachments/assets/3bc96d8a-551c-4dfe-9844-e09aa2cbaf7b" />

3. Observe that:
   - The toggle icon disappears.
   - Console errors appear.

How the toggle button looks on hover when the icon is missing:

<img width="1417" alt="Image" src="https://github.com/user-attachments/assets/9e20b73e-cdf1-476b-93b8-829746df80dd" />

How the toggle button element is rendered in Chrome:

<img width="1263" alt="Image" src="https://github.com/user-attachments/assets/c4891719-9a48-4a7a-b1ee-d944382d6d57" />

---

## Observed Behavior

- On initial load in dark mode, the toggle button displays its icon.
- After switching to light mode:
  - The theme changes correctly.
  - **The toggle button icon disappears.**
  - Console logs show hydration mismatch and DOM-related errors.

---

## Expected Behavior

- The toggle button should display the icon after switching themes.
- The button should remain **visually identifiable**.
- **No hydration or DOM errors** should occur.

### System Info

```bash
System:
  OS: macOS 15.0.1
  CPU: (10) arm64 Apple M1 Pro
  Memory: 103.11 MB / 16.00 GB
[...]

PR #1209, opened 2025-04-24
fix(MainLayout): prevent hydration mismatch by checking isMounted before rendering dark/light mode theme toggle button

gungdekrisna:
### 🔗 Linked issue

#1208 

### ❓ Type of change

### 📚 Description

This PR fixes a hydration mismatch issue that occurs on the [shadcn-vue main page](https://www.shadcn-vue.com/) when switching theme modes.

### 🐛 Problem

When visiting the site with **dark mode enabled by default**, clicking the **theme toggle** button to switch to light mode causes the **icon on the button to disappear**. This leaves the toggle visually invisible and logs the following error in the browser console:
`Hydration completed but contains mismatches.`
`NotFoundError: Failed to execute 'insertBefore' on 'Node': The node before which the new node is to be inserted is not a child of this node.`

This issue is caused by a mismatch between the DOM rendered by the server (SSR) and what Vue renders on the client during hydration.

The `isDark` value is derived from client-only state (like `localStorage` or the user's system preference), which isn't available during SSR. This causes the server to render a different DOM tree than what the client renders once it's mounted, especially for components that rely on `isDark` such as the theme toggle icon.

---

### ✅ Fix

This PR wraps the theme toggle `<Button>` in a `v-if="isMounted"` check to ensure it's only rendered **after the component is mounted on the client**. This avoids any hydration mismatch by skipping this client-only logic during SSR.

```vue
<Button v-if="isMounted" ...>
  <component :is="isDark ? SunIcon : MoonIcon" />
</Button>
```

Before:
before

After:
after

### 📝 Checklist

---

I'm happy to adjust anything if needed — open to any feedback or suggestions! 🙌

### DIFF
--- a/apps/www/.vitepress/theme/layout/MainLayout.vue
+++ b/apps/www/.vitepress/theme/layout/MainLayout.vue
@@ -27,10 +27,13 @@ import ThemePopover from '../components/ThemePopover.vue'
 import { docsConfig } from '../config/docs'
 
 const { radius, theme } = useConfigStore()
+const isMounted = ref(false)
 // Whenever the component is mounted, update the document class list
 onMounted(() => {
   document.documentElement.style.setProperty('--radius', `${radius.value}rem`)
   document.documentElement.classList.add(`theme-${theme.value}`)
+
+  isMounted.value = true
 })
 
 const { frontmatter, isDark } = useData()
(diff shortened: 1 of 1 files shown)


sadeghbarati:
Hi, Thanks for the PR 

How about using the [`ClientOnly`](https://vitepress.dev/reference/runtime-api#clientonly) component here?

gungdekrisna:
> Hi, Thanks for the PR
> 
> How about using the [`ClientOnly`](https://vitepress.dev/reference/runtime-api#clientonly) component here?

Yes, that works too, more concise and clear. Already updated the PR. Thanks for the suggestion!

PR #1220, opened 2025-04-28
Add an option to modify the DialogOverlay class in DialogContent.vue

han1548772930:
### ❓ Type of change

Before:
```js
const props = defineProps<DialogContentProps & { class?: HTMLAttributes['class'] }>()
```
```html
  <DialogOverlay />
```
After:
```js
const props = defineProps<DialogContentProps & { class?: HTMLAttributes['class'], overlayClass?: HTMLAttributes['class'] }>()
```
```html
  <DialogOverlay :class="cn(overlayClass)" />
```

### 📝 Checklist


sadeghbarati:
> This project's goal is to remain aligned with [shadcn/ui](https://github.com/shadcn-ui/ui). If the feature you are requesting is not in the original shadcn/ui, it will not be considered here.

Issue #1308, opened 2025-06-23
[Bug]: devDependencies are not added to package.json


Stawlie:
### Reproduction

Add devDependencies to registry-item and try to add it.

### Describe the bug

devDependencies are not added to package.json

### System Info

```bash
Not necessary
```

### Contributes

- [x] I am willing to submit a PR to fix this issue
- [ ] I am willing to submit a PR with failing tests

sadeghbarati:
Try latest `shadcn-vue` CLI (2.3.1)

PR #1309, opened 2025-06-23
fix(cli): devDependencies update (#1308)

Stawlie:
### 🔗 Linked issue

Issue for devDependencies #1308.

### ❓ Type of change

### 📚 Description

I found that to update devDependencies you need to use addDevDependency function from **nypm**. So i fixed it :)

### 📝 Checklist

### DIFF
--- a/packages/cli/src/utils/updaters/update-dependencies.ts
+++ b/packages/cli/src/utils/updaters/update-dependencies.ts
@@ -1,6 +1,6 @@
 import type { RegistryItem } from '@/src/registry/schema'
 import type { Config } from '@/src/utils/get-config'
-import { addDependency } from 'nypm'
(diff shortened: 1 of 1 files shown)


sadeghbarati:
Hi Thanks for the PR

Can you share what version of `shadcn-vue` you are using?

- https://github.com/unjs/nypm/blob/main/src/api.ts#L96
- https://github.com/unjs/nypm/blob/main/src/api.ts#L158C1-L163C2

Check out these two links please

Also what package should be installed in devDependencies while using `shadcn-vue init/add`?

sadeghbarati:
This should be fixed after this gets merged https://github.com/unovue/shadcn-vue/pull/1330

PR #1330, opened 2025-07-04
feat(cli): add `tsconfig` file detection based on the framework

sadeghbarati:
### ❓ Type of change

### 📚 Description

- ~~Add `tsConfigPath` options to the `components.json` options schema~~
- Check for frameworks for detecting the tsconfig file
- Add main shadcn-ui CLI changes
- Use `c12` again for loading component.json/json5/js/ts  (it is working in Nuxt projects with empty or full components directory)
- Enable `shims` and add nodeProtocol: strip to remove `node:` protocol from the build

### 📝 Checklist


zernonia:
Yo @sadeghbarati ! Thanks for adding this field back into the config. I wonder why this wasn't a request on actual `ui.shadcn`? Perhaps it would be better if we have the same [`monorepo`](https://ui.shadcn.com/docs/monorepo#importing-components) support?

sadeghbarati:
@zernonia 

I think the monorepo CLI code is already here

We might need to remove the tsConfigPath option 

But in Nuxt 4, we have an issue that this option may resolve it  #1340

tsConfigPath is optional 

---

Monorepo is too complicated people is just need to use normal directory for component registry

sadeghbarati:
We have multiple choices

1. make main tsconfig.json/tsconfig.app.json automatically detectable based on framework and framework version (not implemented ❌)
2. add new flag on init and add CLI like --tsconfig-path  (not implemented ❌)
3. `tsConfigPath` in components.json (implemented ✅)

What we should do? which way is more convenient

sadeghbarati:
Hi, @zernonia, could you please check this PR? Thank you ❤️ 

I removed the `tsConfigPath` option and added the tsconfig detection from the existing projectInfos utilities

The PR also has the latest changes from the main shadcn ui too


There is a lot of room for improvement for detection, but let's start small

packages/cli/src/utils/get-project-info.ts
```diff
@@ -41,22 +38,68 @@ const TS_CONFIG_SCHEMA = z.object({
...
+    cwd,
+    deep: 3,
+    ignore: PROJECT_SHARED_IGNORE,
+  })
+
+  // Check for Nuxt
+  if (configFiles.find(file => file.startsWith('nuxt.config.'))) {
+    const isUsingAppDir = await fs.pathExists(path.resolve(cwd, 'app'))
+    return isUsingAppDir ? FRAMEWORKS.nuxt4 : FRAMEWORKS.nuxt3
+  }
```
ZTL-UwU: @sadeghbarati There is a problem with this detection approach. Having the `app` folder does not necessarily mean it is nuxt4, since nuxt3 (with `compatibilityVersion: 4`) also uses the `app` folder structure, and nuxt4 has downward compatibility for the old file structure as well.

Instead, we should probably check the `package.json` for the exact nuxt version using semver, which is what nuxi is doing.

https://github.com/nuxt/cli/blob/77b68177d574e302c91dfc86137e9b9aaa95343f/packages/nuxi/src/utils/versions.ts#L1-L12

```ts
import { readPackageJSON } from 'pkg-types'
import { coerce } from 'semver'

export async function getNuxtVersion(cwd: string, cache = true) {
  const pkg = await readPackageJSON(cwd)
  const pkgDep = pkg?.dependencies?.nuxt || pkg?.devDependencies?.nuxt
  return (pkgDep && coerce(pkgDep)?.version) || '3.0.0'
}
```

ZTL-UwU (changes requested):

packages/cli/src/utils/get-project-info.ts
sadeghbarati: Thanks ❤️

sadeghbarati:
Tests need to be updated. I'm not a pro at it 😭 
Or just merge it without test files, np all good 😁

mayar-deeb-xi:
I came from one of the linked issue..... , any plans to merge this :)

sadeghbarati:
@mayar-deeb-xi  Hi, For now use it like 


1. Based on your package manager for example `pnpm`

```
pnpm add https://pkg.pr.new/shadcn-vue@1330 -D
```

2. Add new script to scripts in `package.json` like


```json
{
    ...
	"scripts": {
		"dev": "vite",
		"shadcn-vue": "shadcn-vue"
	},
	...
}
```

3. run it in your project directory without `pnpm dlx` or `npx`

```
pnpm run shadcn-vue init/add
```

sadeghbarati:
Self-merging for merging these changes to the [`v4-docs`](https://github.com/unovue/shadcn-vue/tree/v4-docs) branch

PR #1367, opened 2025-08-03
docs: document ssrWidth to prevent mobile hydration errors

mathieumagalhaes:
### 🔗 Linked issue
https://github.com/unovue/shadcn-vue/issues/1366

### ❓ Type of change

### 📚 Description

When using any component that relies on useMediaQuery in a Nuxt SSR setup,
a hydration error occurs due to the absence of an ssrWidth value.
https://github.com/unovue/shadcn-vue/issues/1366

**Solution:**
According to the [VueUse documentation](https://vueuse.org/core/useMediaQuery/#server-side-rendering-and-nuxt), the best practice is to provide an ssrWidth to prevent this error. By doing so, we can ensure the component works correctly without affecting other users or breaking existing functionality.

I also ran `lint:fix`

### 📝 Checklist


sadeghbarati (commented):
Hi 👋 

Thanks for the PR, 

Looking to vueuse [`useMediaQuery`](https://github.com/vueuse/vueuse/blob/main/packages/core/useMediaQuery/index.ts#L20) source,
How about adding document about the [`useSSRWidth`](https://vueuse.org/core/useSSRWidth/) so we could prevent duplicated `ssrWidth` option in the code?

for example adding documents here https://www.shadcn-vue.com/docs/installation/nuxt

mathieumagalhaes:
> Hi 👋
> 
> Thanks for the PR,
> 
> Looking to vueuse [`useMediaQuery`](https://github.com/vueuse/vueuse/blob/main/packages/core/useMediaQuery/index.ts#L20) source, How about adding document about the [`useSSRWidth`](https://vueuse.org/core/useSSRWidth/) so we could prevent duplicated `ssrWidth` option in the code?
> 
> for example adding documents here https://www.shadcn-vue.com/docs/installation/nuxt

Hi @sadeghbarati 
Great catch, you're absolutely right. Using provideSSRWidth is a cleaner and more scalable solution moving forward.
I've reverted my previous changes and added the appropriate documentation.

packages/cli/src/utils/get-config.ts
```diff
@@ -4,9 +4,9 @@ import path from 'pathe'
 import { glob } from 'tinyglobby'
 import { z } from 'zod'
 import { getProjectInfo } from '@/src/utils/get-project-info'
+import { logger } from '@/src/utils/logger'
```
mathieumagalhaes: PS. This has been moved by running lint:fix

PR #1398, opened 2025-08-29
fix(module): await component registration

cernymatej:
### ❓ Type of change

### 📚 Description
Since the promises in the component registration were not being awaited, it could cause some components to not be registered for auto-imports.

### 📝 Checklist

### DIFF
--- a/apps/www/src/content/docs/installation/nuxt.md
+++ b/apps/www/src/content/docs/installation/nuxt.md
@@ -87,15 +87,22 @@ import {
 
 export interface ShadcnVueOptions {
   /**
-   * Prefix for all the imported component
+   * Prefix for all the imported component.
+   * @default "Ui"
    */
-  prefix: string;
-
+  prefix?: string
   /**
    * Directory that the component lives in.
-   * @default "~/components/ui"
+   * Will respect the Nuxt aliases.
+   * @link https://nuxt.com/docs/api/nuxt-config#alias
+   * @default "@/components/ui"
    */
-  componentDir: string;
+  componentDir?: string
+  /**
+   * Whether to enable auto-importing of components from the components directory.
+   * @default true
+   */
+  autoImport?: boolean
 }
 
 export default defineNuxtModule<ShadcnVueOptions>({
--- a/apps/www/src/content/docs/installation/nuxt.md
+++ b/apps/www/src/content/docs/installation/nuxt.md
@@ -87,22 +87,15 @@ import {
 
 export interface ShadcnVueOptions {
   /**
-   * Prefix for all the imported component.
-   * @default "Ui"
+   * Prefix for all the imported component
    */
-  prefix?: string
+  prefix: string;
+
   /**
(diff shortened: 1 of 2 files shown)


sadeghbarati:
Hi 👋 

simple way is just don't use `shadcn-nuxt`. right? cause the `shadcn-nuxt` goal is to auto-import the components in the `ui` directory

if you don't want auto-import for `ui` just don't use it

or if you want to disabled auto-import fully use

```ts
export default defineNuxtConfig({
  imports: {
    autoImport: false,
    scan: false,
  }
})
```

or if you want to disable auto-import only for `ui` dir you can use 

```ts
export default defineNuxtConfig({
  hooks: {
    'components:dirs'(dirs) {
      dirs.unshift({
        path: '@/components/ui',
        extensions: [],
      })
    },
  }
})
```

cernymatej:
Hmm, I guess yeah, since it currently doesn't do anything else 😄
But it's probably a good idea to await it, isn't it?

sadeghbarati:
await what line of code can you point it?

cernymatej:
https://github.com/unovue/shadcn-vue/pull/1398/files#diff-e8b52a811377c3181637938504d1ac1400a62288c80a9a73aae897ac6fee3457R64

currently, there's just a foreach, and I think that in some cases, only some of the components will get registered for auto-imports
https://github.com/unovue/shadcn-vue/blob/d90434fcc4e1ee3d96fad9e9952915abaa48a699/packages/module/src/module.ts#L56-L60

(depending on how long other modules take to install)

sadeghbarati:
Right, await must be in the setup scope, could you remove `autoImport` code and replace this block of code, Thanks

<details><summary>Details</summary>
<p>

```ts
try {
  const componentPromises = readdirSync(componentsPath)
    .map(async (dir) => {
      try {
        const filePath = await resolvePath(join(COMPONENT_DIR_PATH, dir, 'index'), { extensions: ['.ts', '.js'] })
        const content = readFileSync(filePath, { encoding: 'utf8' })
        const ast = parseSync(filePath, content, {
          sourceType: 'module',
        })

        const exportedKeys: string[] = ast.program.body
          .filter(node => node.type === 'ExportNamedDeclaration')
          // @ts-expect-error parse return any
          .flatMap(node => node.specifiers?.map(specifier => specifier.exported?.name) || [])
          .filter((key: string) => /^[A-Z]/.test(key))

        exportedKeys.forEach((key) => {
          addComponent({
            name: `${prefix}${key}`, // name of the component to be used in vue templates
            export: key, // (optional) if the component is a named (rather than default) export
            filePath: resolve(filePath),
            priority: 1,
          })
        })
      }
      catch (err) {
        if (err instanceof Error)
          console.warn('Module error: ', err.message)
      }
    })

  await Promise.all(componentPromises)
}
catch (err) {
  if (err instanceof Error)
    console.warn(err.message)
}
```

</p>
</details>

sadeghbarati:
Thank You

Issue #1409, opened 2025-09-11
[Feature]: declare the prop disabled on the Button component


davidfou:
### Describe the feature

Some styling is added when the Button is disabled, but it can be tricky to discover it because the prop `disabled` is not declared.

### Additional information

- [x] I intend to submit a PR for this feature.
- [x] I have already implemented and/or tested this feature.

davidfou:
Apparently, that's not needed.

PR #1410, opened 2025-09-11
feat(button): declare disabled prop

davidfou:
### 🔗 Linked issue

#1409 

### ❓ Type of change

### 📚 Description

Some styling is added on the Button component when it's disabled. Passing the prop works, though, TypeScript considers it as any because it's not declared on the component. This PR declared the prop `disabled` in order to make it discoverable and also ensure it's typed as a boolean

### 📝 Checklist


davidfou:
I'm a bit confused by the diff generated by running the command `pnpm build:registry`, that's why I added to a separate commit. Let me know if I should remove that commit from the branch.

sadeghbarati:
Hi Thanks for the PR,

`inheritAttrs` in vue is **true** by default, there is no need to add disabled prop to the button. instead in the parent or wherever you used the button component pass the disabled prop


Check Vue docs https://vuejs.org/guide/components/attrs

davidfou:
Thanks @sadeghbarati for checking my code. That's indeed what we noticed, passing `:disabled="true"` works as expected. The reason why we'd like to declare it on the component is for discovery and TypeScript validation to ensure a boolean is passed (similar as the native `<button />` HTML element).

I'm new to VueJS, feel free to close this PR in case it's common practice to pass attributes which are not declared on the component.

PR #1545, opened 2025-11-15
fix(docs): correct user-facing imports in component demos

ace-of-aces:
### 🔗 Linked issue

\-

### ❓ Type of change

### 📚 Description

This fixes the internal `@/registry/new-york-v4/...` imports currently being displayed in component demos on the website, which are replaced by the default `@/components/ui/...` import alias. This makes it easier for users to copy-paste code examples without having to manually fix the imports :)

### Details

I copied the `fixImport` function from `~/lib/registry.ts` into the `ComponentSource` component itself, because importing it from `~/lib/registry.ts` caused a 500 error by Vite, related to the `node:fs` import in `registry.ts`:

Screenshot 2025-11-15 at 10 54 24

It might be better here to fix the underlying issue itself, so `registry.ts` can be used in client-side components too. What do you think about this?

Before:
Screenshot 2025-11-15 at 11 20 13

After:
Screenshot 2025-11-15 at 11 19 58

### 📝 Checklist

### DIFF
--- a/apps/v4/components/ComponentSource.vue
+++ b/apps/v4/components/ComponentSource.vue
@@ -16,7 +16,39 @@ const props = withDefaults(defineProps<{
   collapsible: true,
 })
 
-const code = (await import(`@/components/demo/${props.name}.${props.language}?raw`)).default
+/**
+ * @see apps/v4/lib/registry.ts
+ */
+function fixImport(content: string) {
+  // eslint-disable-next-line regexp/no-super-linear-backtracking
+  const regex = /@\/(.+?)\/((?:.*?\/)?(?:components|ui|hooks|lib))\/([\w-]+)/g
+
+  const replacement = (
+    match: string,
+    path: string,
+    type: string,
+    component: string,
+  ) => {
+    if (type.endsWith('components')) {
+      return `@/components/${component}`
+    }
+    else if (type.endsWith('ui')) {
+      return `@/components/ui/${component}`
+    }
+    else if (type.endsWith('hooks')) {
+      return `@/hooks/${component}`
+    }
+    else if (type.endsWith('lib')) {
+      return `@/lib/${component}`
+    }
+
+    return match
+  }
+
+  return content.replace(regex, replacement)
+}
+
+const code = fixImport((await import(`@/components/demo/${props.name}.${props.language}?raw`)).default)
 </script>
 
 <template>
--- a/apps/v4/components/ComponentSource.vue
+++ b/apps/v4/components/ComponentSource.vue
@@ -1,5 +1,6 @@
 <script setup lang="ts">
 import type { HTMLAttributes } from 'vue'
+import { fixImport } from '~/lib/registry-utils'
 import { cn } from '~/lib/utils'
 import { getIconForLanguageExtension } from './Icons'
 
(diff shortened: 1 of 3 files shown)


apps/v4/components/ComponentSource.vue
```diff
@@ -16,7 +16,39 @@ const props = withDefaults(defineProps<{
...
+    component: string,
+  ) => {
+    if (type.endsWith('components')) {
+      return `@/components/${component}`
+    }
+    else if (type.endsWith('ui')) {
+      return `@/components/ui/${component}`
+    }
+    else if (type.endsWith('hooks')) {
+      return `@/hooks/${component}`
```
sadeghbarati:
```suggestion
    else if (type.endsWith('hooks') || type.endsWith('composables')) {
      return `@/composables/${component}`
```

apps/v4/components/ComponentSource.vue
```diff
@@ -16,7 +16,39 @@ const props = withDefaults(defineProps<{
   collapsible: true,
 })
 
-const code = (await import(`@/components/demo/${props.name}.${props.language}?raw`)).default
+/**
+ * @see apps/v4/lib/registry.ts
+ */
+function fixImport(content: string) {
+  // eslint-disable-next-line regexp/no-super-linear-backtracking
+  const regex = /@\/(.+?)\/((?:.*?\/)?(?:components|ui|hooks|lib))\/([\w-]+)/g
```
sadeghbarati: Also `composables` here too beside `hooks`

tnQ

sadeghbarati:
> It might be better here to fix the underlying issue itself, so registry.ts can be used in client-side components too. What do you think about this?

If You have a better solution we are open to consider it :heart:

apps/v4/components/ComponentSource.vue
ace-of-aces: @sadeghbarati Yeah I suspect that's still in there from React, I only copied that from `registry.ts` as stated in my PR description:
> I copied the fixImport function from ~/lib/registry.ts into the ComponentSource component itself, because importing it from ~/lib/registry.ts caused a 500 error by Vite, related to the node:fs import in registry.ts

What do you think: Should we try to fix `registry.ts` so it can be imported in this component and we don't have to keep those functions in sync?

sadeghbarati: One issue that we faced in `ComponentSource.vue` that component is client side only


https://github.com/nuxt/content/blob/2bdb3c76b02839f1de08bf4cdf7272c77397e2c7/src/utils/templates.ts#L128

I don't know it's `Nuxt Content` limitation it can't accept a **Server Component** or **Island Component** in the `markdown` files, 



We tried to make it just like the `component-source.tsx` in shadcn ui but no luck that is why we use simple dynamic import in the component

sadeghbarati: We don't have `use client` or `use server` directives in Nuxt and since we using `Collapsible` component here creating a `Server Component` beside of other `Client Side Components` may take a lot of efforts

ace-of-aces: Ahh okay, I understand..
 I also experimented with solving this another way, but haven't found a solution either.

sadeghbarati (approved):
LGTM, I'll check the `Nuxt Content` limitation later

sadeghbarati:
Seems odd that VitePress markdown syntax has more features than `Nuxt Content` 😬

PR #1554, opened 2025-11-16
style: replace lime with green color palette for green theme

### LINKED ISSUE #1544: Green theme is visually inconsistent
Hey, first of all I'm glad to see that the v4 docs have finally been released!🎉

I noticed the new default theme when visiting the website is the new "Green" theme.

I find the lime color palette in there to not really fit the rest of the shadcn-vue/unovue brand, because the color palette is radically different from everything else.

Also, the charts are still using the regular `green` color palette from Tailwind.

Here's a screenshot of those styles clashing in the Dashboard example:

Image

## My suggestion

Would you consider getting rid of `lime` colors in the green theme and only using the `green` color palette?

As always, I'd be happy to open a PR for this :)

ace-of-aces:
### 🔗 Linked issue

Resolves #1544 

### ❓ Type of change

### 📚 Description

This brings the default green theme back to the `green` color palette instead of the `lime` palette that breaks with the rest of the branding and green colors used for charts.

Just for a more consistent look ❤️ 

Before:

Screenshot 2025-11-16 at 11 52 58
Screenshot 2025-11-16 at 11 53 15

---

After:

Screenshot 2025-11-16 at 11 51 09
Screenshot 2025-11-16 at 11 51 24

### 📝 Checklist

### DIFF
--- a/apps/v4/assets/css/themes.css
+++ b/apps/v4/assets/css/themes.css
@@ -183,25 +183,25 @@
 
 .theme-green .theme-container,
 .theme-green [data-reka-popper-content-wrapper] {
-  --primary: var(--color-lime-600);
-  --primary-foreground: var(--color-lime-50);
-  --ring: var(--color-lime-400);
+  --primary: var(--color-green-600);
+  --primary-foreground: var(--color-green-50);
+  --ring: var(--color-green-400);
   --chart-1: var(--color-green-300);
   --chart-2: var(--color-green-500);
   --chart-3: var(--color-green-600);
   --chart-4: var(--color-green-700);
   --chart-5: var(--color-green-800);
-  --sidebar-primary: var(--color-lime-600);
-  --sidebar-primary-foreground: var(--color-lime-50);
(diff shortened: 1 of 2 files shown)


zernonia:
Thanks for the PR @ace-of-aces , I was thinking of replacing the color as well, but that would cause the color to deviate from the original. Im thinking we can add it as `emerald` color, and keep the current green as is? WDYT? 😁

ace-of-aces:
@zernonia Good point, I only noticed just now that shadcn/ui has the green theme.. (thought this was a shadcn-vue only thing😅)

I'll revert my changes in the green theme then, but also apply the  `lime` palette to the `chart-` variables to have it consistent with the original, if that's okay.

Regarding the extra theme, would you use the tailwind `emerald` palette then, or only use this name for the theme?
If you meant the latter, I think we shouldn't create confusion with the tailwind names, so maybe something different like `mint`?

And should we then set this new theme as the default for users visiting the website? :)

PR #1562, opened 2025-11-18
chore(deps): hoist tsx to root and manage via catalog

jevin98:
If `tsx` is not elevated to the root directory's `package.json`, `tsx` will not be installed in `v4`, causing an error when executing `registry:build`.


sadeghbarati:
Make sense 👍

PR #1563, opened 2025-11-18
chore: use exec instead of execFile for prettier to support Wi…

jevin98:
@sadeghbarati Sorry, I accidentally inserted a Chinese comment earlier.

Since my environment is Windows, when executing the code `https://github.com/jevin98/shadcn-vue/blob/dev/apps/v4/scripts/build-registry.ts#L140`, I used `execFile` to run `prettier`. However, Windows doesn't seem to support `execFile`. Given that my project currently uses pnpm, I've changed it to use `pnpm exec` to run `prettier`.


sadeghbarati:
No problem it's ok 👍 

Marked the commit as chore 🙏

PR #1564, opened 2025-11-18
chore: normalize line endings to LF across platforms

jevin98:
fix [link](https://www.github.com/unovue/shadcn-vue/pull/1560#issuecomment-3545433034)


apps/v4/scripts/build-registry.ts
jevin98: @sadeghbarati And here.

jevin98:
Then all the others are the results of executing `pnpm registry:build-only`.

If you feel this has resulted in too many changes, I can revert to the previous state before `registry:build-only`.

sadeghbarati:
LGTM 🙏

PR #1855, opened 2026-06-16
fix(cli): don't crash on init --preset when cwd has a sibling components/ dir

salmanulfaris:
## Summary

`npx shadcn-vue@latest init --preset <code>` crashes with a misleading **"Invalid components.json configuration"** error on essentially any already-initialized Vue/Nuxt project — reproducible with shadcn-vue's own built-in presets (e.g. `--preset vega`), not specific to any particular preset code.
Issue : https://github.com/unovue/shadcn-vue/issues/1854

```
- Preflight checks.
✔ Preflight checks.
- Verifying framework.

Something went wrong. Please check the error below for more details.

Error:
Invalid components.json configuration in <cwd>.

Message:
Error: Cannot find module 'components'
Require stack:
- <cwd>/components/_index.js

Suggestion:
Check your components.json file for syntax errors or invalid configuration. Run 'npx shadcn@latest init' to regenerate a valid configuration.
```

Plain `init` (no `--preset`) is unaffected — it correctly detects the existing `components.json` and exits early with a clear message.

## Root cause

The `--preset` flow in `commands/init.ts` temporarily renames `components.json` out of the way (`createFileBackup`) so preflight checks can run on what looks like a fresh project. Preflight then calls `getProjectInfo` → `getConfig` → `getRawConfig`, which loads config via:

```ts
loadConfig({ name: 'components', configFile: 'components', cwd, ... })
```

When no `components.json` / `components.config.*` file exists, c12's fallback chain for an **extensionless** `configFile` ends up defaulting `res.configFile` to the literal, unresolved string `"components"`, then calls `existsSync("components")` — relative to the actual **process** `cwd()`, not the `cwd` argument passed into `loadConfig`. Since virtually every Vue/Nuxt project is run from a directory that has a sibling `components/` folder, this accidentally resolves "true" and the bare string `"components"` gets handed to `jiti.import()`, which throws a raw `MODULE_NOT_FOUND` for `'components'` — surfaced from `getRawConfig`'s catch-all as a generic, misleading `ConfigParseError`.

I confirmed this precisely with a minimal standalone repro calling `c12`'s `loadConfig` directly with the exact same options, isolating the trigger to: no `components.json` present + a literal `components/` directory in `process.cwd()` + the explicit `configFile: 'components'` option. Full stack trace and repro steps available on request.

This is an upstream `c12` resolution quirk, but it's straightforward to defend against on our side.

## Fix

In `getRawConfig`'s catch block, detect this specific failure (`error.code === 'MODULE_NOT_FOUND'` and message referencing `'components'`) and treat it the same as "no config file found" (return `null`), rather than surfacing it as a config parse error — since semantically that's exactly what it is.

## Test plan

- [x] Added `test/fixtures/config-none-with-components-dir` (a `package.json` + a `components/` dir, no `components.json`) and a new regression test in `get-config.test.ts`.
- [x] Since the bug is tied to `process.cwd()` rather than the function's `cwd` argument, the test `process.chdir()`s into the fixture for the duration of the assertion.
- [x] Verified the new test **fails** against the pre-fix code with the exact reported error/stack, and **passes** with the fix.
- [x] Full `packages/cli` test suite (440 tests) passes with no regressions.
- [x] `eslint` clean on changed files.

After this fix, `npx shadcn-vue@latest init --preset vega` (or any preset) on an existing project with a `components/` directory should proceed past "Verifying framework" instead of crashing.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CLI config detection to correctly treat “no components config file found” as a non-error, even when a components folder exists, while still surfacing genuine config/import failures.

* **Tests**
  * Added regression coverage for working-directory-related resolution quirks.
  * Added fixtures to ensure broken explicit config imports still raise errors.

### DIFF
--- a/packages/cli/src/utils/get-config.ts
+++ b/packages/cli/src/utils/get-config.ts
@@ -1,4 +1,5 @@
 import type { z } from 'zod'
+import { existsSync } from 'node:fs'
 import { loadConfig } from 'c12'
 import { getTsconfig } from 'get-tsconfig'
 import path from 'pathe'
@@ -166,12 +167,24 @@ export async function getRawConfig(
     // `MODULE_NOT_FOUND` for `'components'` instead of c12 reporting "no
     // config found". This is indistinguishable from "no config file present"
     // from our side, so treat it the same way (return null) rather than
-    // surfacing it as a config parse error.
+    // surfacing it as a config parse error. Only do this when there's no
+    // explicit config file present — if one of these exists, a
+    // MODULE_NOT_FOUND for 'components' is a real failure (e.g. the user's
+    // own components.config.ts importing a missing module) and must still
+    // surface as ConfigParseError instead of being silently swallowed.
     // See: https://github.com/unjs/c12 (upstream issue pending).
+    const hasExplicitConfigFile
+      = existsSync(path.resolve(cwd, 'components.json'))
+        || existsSync(path.resolve(cwd, 'components.config.ts'))
+        || existsSync(path.resolve(cwd, 'components.config.js'))
+        || existsSync(path.resolve(cwd, 'components.config.mjs'))
+        || existsSync(path.resolve(cwd, 'components.config.cjs'))
+
     if (
       error instanceof Error
       && (error as NodeJS.ErrnoException).code === 'MODULE_NOT_FOUND'
       && error.message.includes('\'components\'')
+      && !hasExplicitConfigFile
     ) {
       return null
     }
--- a/packages/cli/src/utils/get-config.ts
+++ b/packages/cli/src/utils/get-config.ts
@@ -155,6 +155,27 @@ export async function getRawConfig(
     return config
   }
   catch (error) {
+    // Work around a c12/jiti resolution bug: when no `components.json` (or
+    // `components.config.*`) file exists, c12's internal fallback chain for
+    // an extensionless `configFile` (here, the literal string "components")
+    // can end up calling `existsSync("components")` relative to the actual
+    // process cwd instead of the resolved project cwd. Since virtually every
+    // Vue/Nuxt project has a sibling `components/` directory, this
+    // accidentally resolves to that directory and gets passed — as a bare,
+    // unresolved specifier — into `jiti.import()`, which then throws a raw
+    // `MODULE_NOT_FOUND` for `'components'` instead of c12 reporting "no
+    // config found". This is indistinguishable from "no config file present"
+    // from our side, so treat it the same way (return null) rather than
+    // surfacing it as a config parse error.
+    // See: https://github.com/unjs/c12 (upstream issue pending).
+    if (
+      error instanceof Error
+      && (error as NodeJS.ErrnoException).code === 'MODULE_NOT_FOUND'
+      && error.message.includes('\'components\'')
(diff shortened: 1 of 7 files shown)


salmanulfaris:
Addressed review feedback: narrowed the `MODULE_NOT_FOUND` guard so it only suppresses the error when **no explicit config file** (`components.json` or `components.config.{ts,js,mjs,cjs}`) exists in `cwd` — matching the actual condition under which the c12/jiti bug occurs.

Added a fixture (`config-explicit-broken-import`) with a real `components.config.ts` that fails to import a module literally named `components`, plus a regression test proving that case still throws `ConfigParseError` and isn't swallowed. Verified the new test fails without the narrowing and passes with it.

Full suite: 441 passed, no regressions.

zernonia (changes requested):
Thanks for the thorough investigation here — the root-cause analysis is spot on. I verified it against c12's source: `res.configFile = tryResolve(...) || source` followed by `existsSync()` on the unresolved `"components"` string (relative to process cwd) is exactly the quirk you described. Two related changes needed before this can merge:

**1. `hasExplicitConfigFile` checks filenames c12 never loads**

Since we pass `configFile: 'components'` explicitly (with `rcFile: false`), c12 resolves `components.<ext>` for each of its supported extensions — it never looks for `components.config.*`. That default filename only applies when `configFile` is omitted. This means:

- The four `components.config.{ts,js,mjs,cjs}` checks are dead code — those files are never loaded as config.
- The real case the guard is meant to protect is unguarded: a user with an actual `components.ts` (or `.mts`, `.yaml`, etc.) config that throws a `MODULE_NOT_FOUND` mentioning `'components'` would now be silently swallowed as "no config found".

c12 exports its extension list, so the check can be derived instead of hand-rolled:

```ts
import { SUPPORTED_EXTENSIONS } from 'c12'

const hasExplicitConfigFile
  = existsSync(path.resolve(cwd, 'components.json'))
    || SUPPORTED_EXTENSIONS.some(ext =>
      existsSync(path.resolve(cwd, `components${ext}`)),
    )
```

**2. The `config-explicit-broken-import` test passes for the wrong reason**

Because c12 never loads `components.config.ts`, the `import 'components'` inside that fixture is never executed. The `ConfigParseError` the test observes is actually the same c12 quirk as the first test (the fixture's `components/` dir triggers it), and it only throws because the guard happens to match the never-loaded `components.config.ts`. Once point 1 is fixed, this test will start failing.

Suggested fix: rename the fixture's config from `components.config.ts` to `components.ts`. Then c12 genuinely loads it, the broken import genuinely throws from inside the user's config, and the test exercises the scenario it describes. While at it, asserting `rejects.toThrowError(ConfigParseError)` instead of the bare `toThrowError()` would make it harder for an unrelated error to satisfy the test.

**Minor:** the comment mentions "upstream issue pending" — it'd be great to actually file it against unjs/c12 and link it there, so future readers can check whether the workaround is still needed.

Everything else looks good — the first regression test is a solid anchor (chdir is genuinely required since the bug keys off process cwd, and it's correctly restored in `finally`), and the quote-delimited `.includes("'components'")` match is well-chosen to avoid catching `'./components'` or `'@/components/x'`.

PR #1842, opened 2026-06-03
fix(cli): use nypm for dependency installs

jevin98:
## Summary

- route CLI dependency installs through `nypm` so pnpm/yarn can use Corepack-aware package manager resolution
- update project creation and MCP dependency installs to use the same install path
- add tests for dependency de-duplication and `pnpm dlx` user-agent fallback

Fixes #1839.

## Verification

- `pnpm --filter shadcn-vue test test/utils/updaters/update-dependencies.test.ts test/utils/create-project.test.ts`
- `pnpm --filter shadcn-vue build`

Notes: `pnpm --filter shadcn-vue typecheck` is currently blocked by existing errors in `src/preset/preset.ts`, `src/registry/api.test.ts`, `src/utils/transformers/transform-icons.ts`, and `src/utils/transformers/transform-import.ts`. The full test suite also still has pre-existing Windows path/local registry test failures unrelated to this change.

## Summary by CodeRabbit

* **Refactor**
  * Switched to a unified package-manager installation layer for dependency installs, improving reliability and consistency across environments.
  * Made project-creation and update installs quieter by default (silent operations) and streamlined the install flow.

* **Tests**
  * Added tests covering dependency management: deduplication, package-manager detection fallback, and empty-list behavior.

Issue #1839, opened 2026-05-31
[Bug]: CLI incompatible with pnpm 11


victpork:
### Reproduction

shadcn-vue CLI exits in the middle when switched to pnpm 11

### Describe the bug

the shadcn-vue@latest script exits in the middle when using pnpm 11:
```
> pnpm dlx shadcn-vue@latest add table
✔ Checking registry.
✔ Updating CSS variables in src/style.css
                                                                                                                                       11:34:33 am
                                                                                                                                       11:34:33 am
Something went wrong. Please check the error below for more details.                                                                   11:34:33 am
If the problem persists, please open an issue on GitHub.                                                                               11:34:33 am
                                                                                                                                       11:34:33 am
Process exited with non-zero status (1)
```
When fallback to pnpm 10, the result is OK:
```
> corepack use pnpm@latest-10
Installing pnpm@10.34.1 in the project...

Lockfile is up to date, resolution step is skipped
Already up to date
Done in 539ms using pnpm v10.34.1

> pnpm dlx shadcn-vue@latest add table
✔ Checking registry.
✔ Updating CSS variables in src/style.css
✔ Installing dependencies.
✔ Created 11 files:
  - src/components/ui/table/Table.vue                                                                                                  11:32:14 am
  - src/components/ui/table/TableBody.vue                                                                                              11:32:14 am
  - src/components/ui/table/TableCaption.vue                                                                                           11:32:14 am
  - src/components/ui/table/TableCell.vue                                                                                              11:32:14 am
  - src/components/ui/table/TableEmpty.vue                                                                                             11:32:14 am
  - src/components/ui/table/TableFooter.vue                                                                                            11:32:14 am
  - src/components/ui/table/TableHead.vue                                                                                              11:32:14 am
  - src/components/ui/table/TableHeader.vue                                                                                            11:32:14 am
  - src/components/ui/table/TableRow.vue                                                                                               11:32:14 am
  - src/components/ui/table/index.ts                                                                                                   11:32:14 am
  - src/components/ui/table/utils.ts
```

### System Info

```bash
System:
    OS: Windows 11 10.0.26200
    CPU: (12) x64 AMD Ryzen 5 5600X 6-Core Processor             
    Memory: 17.42 GB / 31.91 GB
  Binaries:
    Node: 24.16.0 - C:\Program Files\nodejs\node.EXE
    npm: 9.6.2 - C:\Program Files\nodejs\npm.CMD
[...]

PR #1803, opened 2026-05-13
fix(cli): route init dependency installs through detected package manager

zernonia:
### 🔗 Linked issue

Fixes #1801

### ❓ Type of change

- [x] 🐞 Bug fix (a non-breaking change that fixes an issue)

### 📚 Description

`shadcn-vue init` would crash during the registry-dependency install step when the project was already laid out by pnpm but pnpm's lockfile wasn't picked up by detection (e.g. `node_modules/.pnpm/` present but no `pnpm-lock.yaml`). `nypm.addDependency` then fell back to `npm`, which ran lifecycle scripts on pre-existing pnpm-installed packages — failing on fragile prepare scripts (the reporter saw `vp: not found` from `eslint-plugin-oxlint`).

This refactors dependency install to mirror shadcn-ui's pattern:

- New `getPackageManager(cwd, { withFallback })` helper with a clear detection order:
  1. `nypm.detectPackageManager` (lockfile / `packageManager` field)
  2. `node_modules/.pnpm` directory — strong signal pnpm was used even when the lockfile is missing
  3. `npm_config_user_agent` (when `withFallback` is true, e.g. `pnpm dlx`)
  4. Fallback to `npm`
- `updateDependencies` now goes through explicit per-PM install commands (`pnpm add`, `yarn add`, `bun add`, `npm install`, `deno add`) via `tinyexec` instead of `nypm.addDependency`'s opaque wrapper. No more accidental PM switching mid-init.
- `createProject` uses the same helper with `withFallback: true` so `pnpm dlx shadcn-vue init` stays on pnpm even in a parent dir without a lockfile.

The `.pnpm`-dir fallback and `withFallback: true` in `updateDependencies` are deliberately stricter than shadcn-ui — those are the two signals that would have caught #1801.

### 📝 Checklist

- [x] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.

### Test plan

- [x] `pnpm test` — 424 pass, no regressions
- [x] `pnpm lint` clean (only the pre-existing `search.ts` warning)
- [x] Local end-to-end repro of #1801's condition (`node_modules/.pnpm` present, no `pnpm-lock.yaml`) — `updateDependencies` produces `pnpm-lock.yaml`, not `package-lock.json`
- [x] Detection table validated across 13 scenarios: lockfile-present, `.pnpm`-dir-only, user-agent-only, no signals, etc.

## Summary by CodeRabbit

* **New Features**
[...]

Issue #1801, opened 2026-05-12
[Bug]: npm install is failing during pnpm dlx shadcn-vue commands


Soviut:
### Reproduction

NA

### Describe the bug

I ran

```bash
pnpm dlx shadcn-vue@latest init --preset a4rFVmnI --template vite
```
And while part of the process passed...
```
✔ Preflight checks.
✔ Verifying framework. Found Vite.
✔ Validating Tailwind CSS config. Found v4.
✔ Validating import alias.
✔ Writing components.json.
✔ Checking registry.
✔ Updating CSS variables in src/index.css
✔ Updating src/index.css
```
...I got the following error.

```
npm install clsx tailwind-merge shadcn-vue@latest class-variance-authority tw-animate-css reka-ui lucide-vue-next failed.

//// I REMOVED SEVERAL DEPRECATION WARNINGS HERE ////

npm error code 127
npm error path /home/ian/projects/collector/node_modules/.pnpm/eslint-plugin-oxlint@1.60.0_oxlint@1.60.0/node_modules/eslint-plugin-oxlint
npm error command failed
npm error command sh -c vp config
npm error sh: 1: vp: not found
npm error A complete log of this run can be found in: /home/ian/.npm/_logs/2026-05-12T04_47_32_832Z-debug-0.log
```

**NOTE:** the `error path` seems to change since it is now giving me errors about jsdom instead of oxlint.

### System Info

```bash
System:
    OS: Linux 6.8 Ubuntu 24.04.3 LTS 24.04.3 LTS (Noble Numbat)
    CPU: (16) x64 13th Gen Intel(R) Core(TM) i5-1340P
    Memory: 23.62 GB / 31.06 GB
    Container: Yes
    Shell: 5.2.21 - /bin/bash
  Binaries:
    Node: 22.14.0 - /home/ian/.nvm/versions/node/v22.14.0/bin/node
    npm: 10.9.2 - /home/ian/.nvm/versions/node/v22.14.0/bin/npm
    pnpm: 10.33.0 - /home/ian/.nvm/versions/node/v22.14.0/bin/pnpm
    bun: 1.1.38 - /home/ian/.bun/bin/bun
  Browsers:
    Chrome: 147.0.7727.55
    Firefox: 150.0.2
    Firefox Developer Edition: 150.0.2
  npmPackages:
    vue: ^3.5.34 => 3.5.34
```

### Contributes

- [ ] I am willing to submit a PR to fix this issue
[...]

PR #1875, opened 2026-07-10
fix(cli): use `execa` for dependency installs

### LINKED ISSUE #1873: [Bug]: pnpm dlx not working on pnpm 11
### Reproduction

https://stackblitz.com/edit/vitejs-vite-ko6wxvyq?file=index.html&terminal=dev

### Describe the bug

When I ran  `pnpm dlx shadcn-vue@latest init --preset a2QzT0aY --template vite` it throw error as below.

this also happen with some component while some are wokring.

```
pnpm dlx shadcn-vue@latest add card
❯ pnpm dlx shadcn-vue@latest add card
✔ Checking registry.
✔ Updating CSS variables in app/frontend/entrypoints/application.css
✔ The file card/Card.vue already exists. Would you like to overwrite? … yes
✔ The file card/CardAction.vue already exists. Would you like to overwrite? … no
? The file card/CardContent.vue already exists. Would you like to overwrite? › (y/N)


```

```
pnpm dlx shadcn-vue@latest add button
✔ Checking registry.
✔ Updating CSS variables in app/frontend/entrypoints/application.css
⠇ Installing dependencies. ERR_PNPM_UNEXPECTED_STORE  Unexpected store location

The dependencies at "/home/user/Documents/project-1/node_modules" are currently linked from the store at "/home/user/.local/share/pnpm/store/v11".

pnpm now wants to use the store at "/home/user/.local/share/pnpm/store/v10" to link dependencies.

If you want to use the new store location, reinstall your dependencies with "pnpm install".

You may change the global store location by running "pnpm config set store-dir <dir> --global".
(This error may happen if the node_modules was installed with a different major version of pnpm)
                                                                                                                                                                                    6:25:28 PM
                                                                                                                                                                                    6:25:28 PM
Something went wrong. Please check the error below for more details.                                                                                                                6:25:28 PM
If the problem persists, please open an issue on GitHub.                                                                                                                            6:25:28 PM
                                                                                                                                                                                    6:25:28 PM
corepack pnpm add reka-ui @vueuse/core @lucide/vue failed.    
```

### System Info

```bash
System:
    OS: Linux 6.12 Pop!_OS 22.04 LTS
    CPU: (14) x64 Intel(R) Core(TM) Ultra 5 125U
    Memory: 1.77 GB / 15.13 GB
    Container: Yes
    Shell: 5.8.1 - /usr/bin/zsh
  Binaries:
    Node: 24.15.0 - /home/user/.local/share/mise/installs/node/24/bin/node
    Yarn: 4.9.2 - /home/user/.local/share/mise/installs/yarn/latest/bin/yarn
    npm: 11.12.1 - /home/user/.local/share/mise/installs/node/24/bin/npm
    pnpm: 11.3.0 - /home/user/.local/share/pnpm/bin/pnpm
    Deno: 2.7.12 - /home/user/.local/share/mise/installs/deno/2.7.12/bin/deno
  Browsers:
    Brave Browser: 150.1.92.134
    Firefox: 150.0.3
    Firefox Developer Edition: 150.0.3
  npmPackages:
    @vueuse/core: ^14.3.0 => 14.3.0 
    reka-ui: ^2.10.1 => 2.10.1 
    shadcn-vue: ^2.7.4 => 2.7.4 
    vue: ^3.5.38 => 3.5.38
```

### Contributes

- [ ] I am willing to submit a PR to fix this issue

jevin98:
## Summary

- replace `nypm` install helpers with explicit package-manager commands executed through `execa`
- use the same direct install path for registry dependencies, project scaffolding, and MCP setup
- keep the existing `.pnpm` directory and `npm_config_user_agent` detection fallbacks
- add `execa@^9.6.0` and update the related tests and lockfile

Fixes #1873.

## Historical context and root cause

This divergence from shadcn/ui was intentional rather than an omitted sync:

1. #1803 mirrored shadcn/ui's explicit per-package-manager commands, but used the repository-standard `tinyexec`. It also retained stricter detection fallbacks to prevent `nypm` from silently falling back to npm (#1801).
2. #1839 reported that the direct install path failed on Windows with Node 24 and pnpm 11.
3. #1842 consequently routed dependency installation through `nypm` so it could use Corepack-aware resolution.
4. #1873 exposed the downside of that workaround: Corepack can select pnpm 10 for a project whose `node_modules` was created with pnpm 11, causing `ERR_PNPM_UNEXPECTED_STORE`.

Current shadcn/ui uses `execa` to invoke the detected package manager directly in both dependency updates and MCP installation:

- https://github.com/shadcn-ui/ui/blob/main/packages/shadcn/src/utils/updaters/update-dependencies.ts
- https://github.com/shadcn-ui/ui/blob/main/packages/shadcn/src/commands/mcp.ts

This PR aligns the install execution path with shadcn/ui while preserving shadcn-vue's stricter package-manager detection from #1803.

## Focused reproduction

A focused Windows reproduction was run with the exact versions reported by #1839:

- Windows
- Node 24.16.0
- pnpm 11.5.0
- `execa@9.6.0`
- `tinyexec@1.0.2`

Both runners successfully resolved pnpm 11.5.0 and completed an offline local `pnpm add`. This indicates that direct process execution is not inherently incompatible with pnpm 11; the currently reproducible failure is the Corepack version selection introduced by the `nypm` install path.

## Validation

- `pnpm --filter shadcn-vue test test/utils/updaters/update-dependencies.test.ts test/utils/create-project.test.ts test/commands/init.test.ts`
  - 32 passed, 1 skipped
- changed-file ESLint
- `pnpm install --filter shadcn-vue --offline --frozen-lockfile --ignore-scripts`
- `pnpm --filter shadcn-vue build`
- `git diff --check`

The full CLI test run still has 10 existing Windows path/local-file failures (431 tests pass), and `pnpm --filter shadcn-vue typecheck` still reports existing errors in preset and transformer files. None of those failing files are changed by this PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved dependency installation for project creation, MCP setup, and dependency updates across npm, pnpm, and Deno, including proper handling of development dependencies.
  * Added correct Deno dependency spec rewriting (e.g., `npm:<name>`) during MCP initialization.
  * Refined package-manager fallback behavior to choose the right install command when detection isn’t available.
* **Tests**
  * Updated existing tests to assert execution via the external install command runner and correct `cwd` usage.
  * Added a new test suite covering `mcp init` on Deno dependency prefixing and invocation.

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -89,6 +89,7 @@
     "dedent": "^1.7.0",
     "deepmerge": "^4.3.1",
     "diff": "^8.0.2",
+    "execa": "^9.6.0",
     "fs-extra": "^11.3.2",
     "fuzzysort": "^3.1.0",
     "get-tsconfig": "^4.13.0",
--- a/packages/cli/src/commands/mcp.ts
+++ b/packages/cli/src/commands/mcp.ts
@@ -2,8 +2,8 @@ import { promises as fs } from 'node:fs'
 import { StdioServerTransport } from '@modelcontextprotocol/sdk/server/stdio.js'
 import { Command } from 'commander'
 import deepmerge from 'deepmerge'
+import { execa } from 'execa'
 import fsExtra from 'fs-extra'
-import { addDevDependency } from 'nypm'
 import path from 'pathe'
 import prompts from 'prompts'
 import z from 'zod'
@@ -251,10 +251,13 @@ async function runMcpInit(options: z.infer<typeof mcpInitOptionsSchema>) {
 }
 
 async function installMcpDependencies(cwd: string) {
-  const packageManager = await getPackageManager(cwd, { withFallback: true })
-  await addDevDependency(DEPENDENCIES, {
+  const packageManager = await getPackageManager(cwd, {
+    withFallback: true,
+  })
+  const installCommand = packageManager === 'npm' ? 'install' : 'add'
+  const devFlag = packageManager === 'npm' ? '--save-dev' : '-D'
+
+  await execa(packageManager, [installCommand, devFlag, ...DEPENDENCIES], {
     cwd,
-    packageManager,
-    silent: true,
   })
 }
--- a/packages/cli/src/commands/mcp.ts
+++ b/packages/cli/src/commands/mcp.ts
@@ -256,8 +256,11 @@ async function installMcpDependencies(cwd: string) {
   })
   const installCommand = packageManager === 'npm' ? 'install' : 'add'
   const devFlag = packageManager === 'npm' ? '--save-dev' : '-D'
+  const dependencies = packageManager === 'deno'
+    ? DEPENDENCIES.map(dependency => `npm:${dependency}`)
+    : DEPENDENCIES
 
-  await execa(packageManager, [installCommand, devFlag, ...DEPENDENCIES], {
+  await execa(packageManager, [installCommand, devFlag, ...dependencies], {
     cwd,
   })
 }
--- a/packages/cli/src/utils/create-project.ts
+++ b/packages/cli/src/utils/create-project.ts
@@ -1,8 +1,8 @@
 import type { z } from 'zod'
 import type { initOptionsSchema } from '@/src/commands/init'
+import { execa } from 'execa'
 import fs from 'fs-extra'
 import { downloadTemplate } from 'giget'
-import { installDependencies } from 'nypm'
 import path from 'pathe'
 import prompts from 'prompts'
 import { x } from 'tinyexec'
@@ -134,11 +134,7 @@ export async function createProject(
     })
 
     // Install dependencies
-    await installDependencies({
-      cwd: projectPath,
-      packageManager,
-      silent: true,
-    })
+    await execa(packageManager, ['install'], { cwd: projectPath })
 
     createSpinner?.succeed(`Created a new ${template} project.`)
   }
--- a/packages/cli/src/utils/updaters/update-dependencies.ts
+++ b/packages/cli/src/utils/updaters/update-dependencies.ts
@@ -3,7 +3,8 @@ import type { RegistryItem } from '@/src/schema'
 import type { Config } from '@/src/utils/get-config'
 import fs from 'node:fs'
 import path from 'node:path'
-import { addDependency, addDevDependency, detectPackageManager } from 'nypm'
+import { execa } from 'execa'
+import { detectPackageManager } from 'nypm'
 import { spinner } from '@/src/utils/spinner'
 
 export type SupportedPackageManager = PackageManagerName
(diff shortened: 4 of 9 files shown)


sadeghbarati:
Hi, did you try the latest version of these packages?

`nypm` latest version is [0.6.8](https://github.com/unjs/nypm/releases/tag/v0.6.8) ours is 0.6.2
`tinyexec` is [1.2.4](https://github.com/tinylibs/tinyexec/releases/tag/1.2.4) ours is 1.0.2


for pnpm 11, can you confirm if the issue will be gone if the user set `packageManager` field in `package.json` 
and also with `corepack enable`

jevin98:
Hi, I checked this with the newer versions and the Corepack behavior in mind.

Adding an exact `"packageManager": "pnpm@11.x"` field does avoid the store mismatch, because Corepack will use that pnpm version. However, `corepack enable` alone does not change the selected version. In #1873, the failing command was already `corepack pnpm add ...`, so Corepack was available; the problem was that it selected pnpm 10 when the project had been installed with pnpm 11.

I also checked `nypm@0.6.8`. Upgrading alone would not change this behavior because nypm still uses Corepack by default. It now supports `corepack: false`, which could be another option if we want nypm to call the pnpm executable from PATH directly. A newer `tinyexec` may help with the older Windows execution issue in #1839, but it would not affect the pnpm version selected by Corepack in #1873.

My concern with relying on the `packageManager` field is that it turns this into a project prerequisite. shadcn-vue already supports `pnpm dlx` and projects without that field through the user-agent fallback, so those projects should still use the package manager that invoked the CLI without requiring users to update their project metadata first.

The current shadcn/ui CLI invokes the detected package manager directly through `execa`, which is the behavior this PR follows while keeping shadcn-vue's additional `.pnpm` directory and user-agent detection.

I ran a focused Windows check with Node 24.16.0 and pnpm 11.5.0 as well. Both direct `execa` and `tinyexec` execution resolved pnpm 11.5.0 and completed a local install successfully.

If you would prefer to keep nypm, I can also test `nypm@0.6.8` with `corepack: false` against the full #1839 reproduction. But upgrading the packages and requiring users to set `packageManager` would only be a workaround for configured projects, rather than the same general fix.

sadeghbarati:
https://github.com/unovue/shadcn-vue/pull/1878

Here a example PR for testing the CLI without `corepack` and `packageManager`
Seems working for me locally

```
https://pkg.pr.new/shadcn-vue@1878
```

---

Here the ready StackBlitz playground for you to test, added the `pkg.pr.new` to the devDeps and you need to run it with `package.json` scripts

like `pnpm shadcn-vue init/add`

https://stackblitz.com/edit/vitejs-vite-tys3sf5m?file=package.json


---


Btw we also have `tinyexec` which is lighter version of `execa`

sadeghbarati:
Thanks for the PR, closing

Issue #1898, opened 2026-07-26
Static audit of apps/v4: a few accessibility and metadata findings


vinayakkulkarni:
## Summary

I built a static auditor for shadcn-vue applications and ran it against `apps/v4` in this repo. Sharing the results because a handful of the findings look genuinely actionable, and because the tool may be useful to people building on shadcn-vue.

**[shadscan-vue](https://github.com/vinayakkulkarni/shadscan-vue)** — a Vue port of [TheOrcDev/shadscan](https://github.com/TheOrcDev/shadscan), built with the original author's permission. 62 deterministic rules across foundation, interaction, states, accessibility, forms, and production polish. No network calls, no telemetry, no model calls — it reads files and reports line numbers.

```bash
npx shadscan-vue
```

## About the score

`apps/v4` scores **61/100 (D)**.

Running this against your repo immediately exposed a bug in my tool rather than in your code: the first pass produced 539 findings, 78% of which pointed at `components/demo/**`, `components/_internal/sink/**`, and generated registry output. Those files are *supposed* to be minimal — a demo showing `<Button size="icon"><TrashIcon /></Button>` is demonstrating an API, not shipping an unlabelled control.

I fixed that before opening this issue. Illustrative directories are now excluded the same way `*.stories.*` files already were, which took the report from 539 findings to 132, of which **93% are real application code**. Everything below is from `shadscan-vue@0.3.0` and is reproducible with the command above.

## The findings that look real

These are outside demo and generated directories, and I verified each one by reading the source rather than trusting the scanner:

**1. Icons inside controls are not hidden from screen readers**

`@lucide/vue` does not set `aria-hidden` on its output — I checked the published package, there is no `aria-hidden` anywhere in it. So an icon inside a labelled control gets announced as an extra node. (For contrast, `@nuxt/icon` applies `aria-hidden: true` by default, which is why the same pattern is fine in Nuxt-icon codebases.)

Example — `components/Announcement.vue:10`:
```vue
<NuxtLink href="/docs/components/message-scroller">
  Introducing Chat Components <ArrowRightIcon />
</NuxtLink>
```
17 occurrences in application code. `<ArrowRightIcon aria-hidden="true" />` resolves it.

**2. `SelectTrigger` without an accessible name**

`components/ColorFormatSelector.vue:31` — the `SelectTrigger` has no `aria-label` and no associated `<Label>`. Since the trigger is the focusable control (not the `Select` root), a screen reader announces only the placeholder.

[...]

PR #1900, opened 2026-07-26
ci: add an advisory UI audit for apps/v4

vinayakkulkarni:
Follows up on #1898, and implements option 1 from that issue — **advisory only, nothing gated**.

*(Replaces #1899, which I closed after a force-push from a shallow clone detached the branch history and blew its diff up to the whole tree. Same change, now correctly based on `dev`.)*

## What this adds

A single workflow, `.github/workflows/ui-audit.yaml`, that runs [shadscan-vue](https://github.com/vinayakkulkarni/shadscan-vue) against `apps/v4` on pull requests touching it and writes the score, category table, and every failing rule to the job summary.

```yaml
- uses: vinayakkulkarni/shadscan-vue@v0
  with:
    path: apps/v4
```

That is the whole change: 31 lines, one file, no dependency added to the repository.

## Deliberately non-blocking

The job sets `continue-on-error: true`, so it can never fail a contributor's pull request. It reports and stops there. If you later decide you want a floor, `fail-under: 61` would pin today's score and catch regressions — but I have left that out on purpose, since gating is your call and not mine to assume.

Path filters mirror `test.yaml`, so it only runs when `apps/v4` changes.

## No install step

shadscan-vue reads source files and never starts the app, so the audit runs on a bare checkout. I verified this against a fresh clone of `dev` with no `node_modules` present: complete coverage, 204 files, **61/100 (D)** — matching the number in #1898 exactly.

## Verified before proposing

I ran this workflow on a fork first rather than opening it here untested, and that surfaced a rendering bug in my own action: rule messages name the offending element, so `Routable surface starts at <h2> with no <h1>` was rendered as a literal heading and broke the summary apart. Fixed in shadscan-vue 0.3.1 by escaping messages before they reach the renderer; `@v0` already points at it.

## A note on `actions/checkout@v7`

Your other workflows pin `@v4`, which now runs on Node 20 and logs a deprecation warning on every run. Rather than inherit that in a new file, this uses `@v7`. Happy to change it to `@v4` if you would prefer the pin stay uniform across workflows and handle the bump separately.

## Easy to decline

If this is not something you want in CI, close it — no hard feelings, and #1898 stands on its own as a list of findings you can act on or ignore. I would rather you have the option than assume you want the tool.

Disclosure: I wrote shadscan-vue.

## Summary by CodeRabbit

* **Chores**
  * Added automated UI accessibility and quality audits for applicable pull requests.
  * Audit results are reported for review without blocking changes from being merged.

### DIFF
--- a/.github/workflows/ui-audit.yaml
+++ b/.github/workflows/ui-audit.yaml
@@ -1,0 +1,31 @@
+name: UI Audit
+
+on:
+  pull_request:
+    branches:
+      - dev
+    paths:
+      - apps/v4/**
+      - .github/workflows/ui-audit.yaml
+
+permissions:
+  contents: read
+
+jobs:
+  audit:
+    name: shadscan-vue
+    runs-on: ubuntu-latest
+    # Advisory only. The audit reports to the job summary and never blocks a
+    # pull request, so a regression is visible without gating contributors.
+    continue-on-error: true
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v7
+
+      # shadscan-vue reads source files and never runs the app, so the audit
+      # needs no install step and adds no dependency to this repository.
+      - name: Audit apps/v4
+        uses: vinayakkulkarni/shadscan-vue@v0
+        with:
+          path: apps/v4


zernonia:
Thanks for the effort here, and genuinely for the framing — the disclosure up front and the "easy to decline" note are appreciated.

We're going to pass on this one.

The main reason isn't the 31 lines, it's what they pull in. `@v0` is a mutable tag, and the action shells out to `npx --yes shadscan-vue@latest`, so what executes in our CI can change at any time with no diff in this repo. That's a standing surface we'd be taking on for a report that is non-blocking by design and lives in the job summary — in practice nobody would open it. Adding a job to most `apps/v4` PRs for output that's always green isn't a trade we want.

The findings in #1898 are a different story, and that part was useful. The missing `aria-hidden` on `@lucide/vue` icons inside controls and the missing `autocomplete` on the account form inputs are real, and we'll pick those up. If you'd rather send those as a PR yourself, that's welcome — otherwise we'll handle it.

Thanks again, and good luck with shadscan-vue.

PR #1912, opened 2026-08-01
feat(cli): resolve registry items from public GitHub repositories

### LINKED ISSUE #1911: [Feature]: Resolve registry items from a public GitHub repository (owner/repo/item)
### Describe the feature

The React CLI can install a registry item straight from a public GitHub repository, with no configuration at all:

```bash
npx shadcn@latest add evilrabbit/lifeline/personal
```

`shadcn-vue` has no equivalent. `packages/cli/src/registry/parser.ts` recognises exactly one non-URL form — `@namespace/item` — so anything without a leading `@` falls through as a bare item name and is looked up in the default shadcn-vue registry. `owner/repo/item` therefore fails rather than resolving.

**Why it's worth having.** Publishing a registry today means building `r/*.json`, hosting it somewhere, and asking every consumer to add a `registries` entry to their `components.json`. That is a lot of ceremony for someone who just wants to share a component from a repo they already have. It also means a registry that works with the React CLI does not work here, which is a papercut for anyone porting one.

**Proposed behaviour** — a third specifier alongside the existing two:

| Form | Resolution |
| --- | --- |
| `https://…/r/item.json` | fetch directly *(exists today)* |
| `@ns/item` | look `@ns` up in `components.json` → `registries` *(exists today)* |
| `owner/repo/item` | **new** — read `registry.json` from the repository root |
| `owner/repo/item#ref` | **new** — pin a branch, tag, or commit |

Resolution would follow the reference implementation in `shadcn-ui/ui` (`packages/shadcn/src/registry/{address,github,github-ref,source}.ts`):

1. Validate owner and repo strictly, before any network call, so a bare item name containing slashes is never mistaken for a GitHub address.
2. Resolve the ref — explicit `#ref` if given, otherwise the repository's default branch. Not hardcoded to `main`; plenty of registries still sit on `master`, and guessing produces a confusing 404.
3. `GET https://raw.githubusercontent.com/{owner}/{repo}/{sha}/registry.json`, parse against `registrySchema`, find the item by name.
4. Fetch each `files[].path` from the same commit, so a branch that moves mid-install cannot produce a half-updated component.
5. `registryDependencies` keep their existing meaning — a full URL is fetched as-is, a bare name resolves against the default registry.

**Worth deciding explicitly:**

- **Private repos.** `raw.githubusercontent.com` will not serve them. This should fail with a message that says so, rather than a bare 404.
- **Security.** A registry item can write a file anywhere its `target` says, and `resolveFilePath` in `update-files.ts` joins `target` onto the project root without a traversal check. That is a bigger deal for a zero-configuration form than for a URL someone pasted deliberately, so absolute and `..`-traversing values in both `path` and `target` should be rejected for this scheme.
- **Nested names.** `owner/repo/forms/login` should mean the item *named* `forms/login` in the root `registry.json`, not a path to a nested `registry.json` — matching the React CLI. Worth documenting either way.
- **Rate limits.** Unauthenticated raw requests are limited, so the registry.json and the resolved ref want caching per `(owner, repo, ref)`, and item file reads want a concurrency cap.

I have this implemented and tested against the reference behaviour, and I'm happy to open the PR — but I wanted to check first whether you'd want parity with the React CLI here or a different resolution scheme.

### Additional information

- [x] I intend to submit a PR for this feature.
- [x] I have already implemented and/or tested this feature.

peter-lewis:
### 🔗 Linked issue

Resolves #1911

### ❓ Type of change

### 📚 Description

Adds a third `add` specifier alongside URLs and `@namespace/item`, so a registry in a public GitHub repository installs with no configuration at all:

```bash
npx shadcn-vue@latest add owner/repo/item
npx shadcn-vue@latest add owner/repo/item#v1.0.0
npx shadcn-vue@latest search owner/repo          # list what a repo publishes
```

The CLI resolves the repository's default branch, reads `registry.json` from the root over `raw.githubusercontent.com`, finds the item by name, and fetches its files from that same commit. This mirrors the resolution scheme in `shadcn` (`packages/shadcn/src/registry/{address,github,github-ref,source}.ts`), so a registry that works with the React CLI works here unchanged.

**New files** (`packages/cli/src/registry/`)

| File | |
| --- | --- |
| `address.ts` | classifies a specifier; validates owner/repo/ref before any network call |
| `github-ref.ts` | resolves a ref to a commit SHA via `git ls-remote`, cached per repository |
| `source.ts` | reads a `registry.json` through a reader abstraction and inlines item file content |
| `github.ts` | the `raw.githubusercontent.com` reader |
| `proxy.ts` | the shared `undici` agent, lifted out of `fetcher.ts` so both fetchers use one |

**Wired into** `resolver.ts` (routing), `api.ts` (`getRegistry` accepts `owner/repo`), `errors.ts` (three new error types), `index.ts` (exports).

#### Notes for review

- **The default branch is resolved, never assumed.** `git ls-remote --symref` gives us `HEAD`, so registries on `master` work. This does mean Git is required for this scheme — it is also what keeps us off the GitHub API and its 60-requests/hour unauthenticated limit. Branches win over tags when a name is ambiguous, and annotated tags resolve to the commit they point at.
- **Every file of an item comes from one commit.** The ref is resolved once per reader, so a branch that moves mid-install cannot produce a half-updated component.
- **Path traversal is rejected.** `resolveFilePath` in `update-files.ts` joins `files[].target` onto the project root with no traversal check. For a URL someone pasted deliberately that is their call, but this form takes no configuration from the user, so absolute and `..`-traversing values in both `path` and `target` are rejected for source-loaded items.
- **Caching.** Resolved refs are kept for the process (a `git ls-remote` per repo is the expensive part); file bodies are cached by resolved raw URL and honour `useCache`. Item file reads are capped at 8 concurrent requests.
- **`registryDependencies` are unchanged.** A full URL is fetched as-is; a bare name resolves against the default registry, exactly as for any other item. Matching the React CLI here rather than resolving bare names within the same repo.
- **Nested item names.** `owner/repo/forms/login` is the item *named* `forms/login` in the root `registry.json`, not a path to a nested one. Documented.
- **Not included:** `validateGitHubRegistrySource` from the reference, since `shadcn-vue` has no `registry validate` command to call it. Easy to add if one lands.

#### Testing

66 new tests across `address.test.ts`, `github-ref.test.ts`, and `github.test.ts` — the last drives the whole path end to end with `msw`, covering ref pinning, a missing root `registry.json`, an unknown item name, duplicate item names, and traversal rejection.

Also verified live against a real registry (39 files, two-level `registryDependencies` chain, no `components.json` entry):

```bash
npx shadcn-vue add peter-lewis/lifeline/personal
```

and confirmed the existing `@namespace/item` and URL forms still install correctly.

`pnpm test` passes from the root. No dependency or lockfile changes.

### 📝 Checklist

## Summary by CodeRabbit

* **New Features**
  * Added support for installing registry components directly from public GitHub repositories.
  * Added optional branch, tag, or commit references, including automatic default-branch resolution.
  * Added support for GitHub catalogs, dependencies, namespaced items, and direct component paths.
  * Added validation for repository references, file paths, registry contents, and missing items with clearer error guidance.

* **Documentation**
  * Expanded CLI and registry documentation with GitHub installation formats, requirements, dependency handling, and safety restrictions.

### DIFF
--- a/apps/v4/content/docs/06.cli.md
+++ b/apps/v4/content/docs/06.cli.md
@@ -67,6 +67,20 @@ Use the `add` command to add components and dependencies to your project.
 npx shadcn-vue@latest add [component]
 ```
 
+A component can be named in any of these ways:
+
+| Form | Resolves to |
+| --- | --- |
+| `button` | an item in the default `shadcn-vue` registry |
+| `@acme/button` | an item in a registry configured under `registries` in `components.json` |
+| `https://acme.com/r/button.json` | a registry item fetched directly from that URL |
+| `./button.json` | a registry item on disk |
+| `owner/repo/button` | an item in the root `registry.json` of a public GitHub repository |
+| `owner/repo/button#v1.0.0` | the same, pinned to a branch, tag or commit |
+
+See [Install from GitHub](/docs/registry/getting-started#install-from-github) for
+the GitHub form.
+
 **Options**
 
 ```bash
--- a/apps/v4/content/docs/registry/getting-started.md
+++ b/apps/v4/content/docs/registry/getting-started.md
@@ -239,6 +239,24 @@ the commit it points at.
 npx shadcn-vue@latest search owner/repo
 ```
 
+### Only install from repositories you trust
+
+Adding a registry item runs third-party code on your machine. This is true of
+every registry — hosted, local or GitHub — but the GitHub form is worth calling
+out because it needs no configuration, so a single command from a README is
+enough to install from a repository you have never looked at.
+
+A registry item can list npm `dependencies`, which the CLI installs with your
+package manager, which in turn runs that package's install scripts. It can also
+write files into your project. Treat `npx shadcn-vue@latest add owner/repo/item`
+with the same care as `npm install owner-repo` — read the registry first if you
+do not know who owns it.
+
+The CLI does reject `path` and `target` values that are absolute or use `..` to
+escape the project, both for the item you asked for and at the point files are
+written. That is a guard against mistakes and a hostile item's easiest trick; it
+is not a sandbox.
+
 ### Notes
 
 - **Item names may contain slashes.** `owner/repo/forms/login` installs the item
@@ -250,10 +268,9 @@ npx shadcn-vue@latest search owner/repo
   which is what lets it find the default branch instead of guessing `main`.
 - **`registryDependencies` are resolved the usual way.** A full URL is fetched
   as-is; a bare name such as `button` resolves against the default `shadcn-vue`
-  registry, exactly as it does for any other registry item.
-- **File paths and targets must stay inside the project.** Because this form
-  takes no configuration from the user, the CLI rejects any `path` or `target`
-  that is absolute or uses `..` to escape.
+  registry, exactly as it does for any other registry item. Note that a
+  dependency can point anywhere, so trusting a repository means trusting what it
+  depends on too.
 
 If you would rather not type the repository on every command, register a
 namespace in `components.json` and install by that instead. A namespace maps to
--- a/apps/v4/content/docs/registry/getting-started.md
+++ b/apps/v4/content/docs/registry/getting-started.md
@@ -256,7 +256,9 @@ npx shadcn-vue@latest search owner/repo
   that is absolute or uses `..` to escape.
 
 If you would rather not type the repository on every command, register a
-namespace in `components.json` and install by that instead:
+namespace in `components.json` and install by that instead. A namespace maps to
+a hosted registry URL template — not to a GitHub repository — so this is an
+alternative to the GitHub form above rather than a shorthand for it:
 
 ```json title="components.json"
 {
--- a/apps/v4/content/docs/registry/getting-started.md
+++ b/apps/v4/content/docs/registry/getting-started.md
@@ -204,3 +204,68 @@ To install a registry item using the `shadcn-vue` CLI, use the `add` command fol
 ```bash
 npx shadcn-vue@latest add http://localhost:3000/r/hello-world.json
 ```
+
+## Install from GitHub
+
+If your registry lives in a **public** GitHub repository with a `registry.json`
+at its root, you do not have to build, host or configure anything. The CLI can
+read it directly:
+
+```bash
+npx shadcn-vue@latest add owner/repo/hello-world
+```
+
+The CLI resolves the repository's default branch, reads `registry.json` from the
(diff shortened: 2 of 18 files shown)


peter-lewis:
Thanks for the review — pushed d624251. Taken and skipped, with reasoning:

**Fixed**

- **Rejected promise stays in `fileCache`** — real bug, and the best catch of the review. One transient failure would have been replayed for the rest of the run. Now evicted on rejection, matching what `resolveGitHubRef` already did. Regression test added; I verified it fails without the fix.

  Writing that test turned up something worth knowing: my first attempt used a 500 and *passed* without the fix, because `ofetch` retries `retryStatusCodes` on GET and silently succeeded on the retry. The test now uses a 404, which isn't retried.

- **No timeout on the raw fetch** — correct, nothing else bounded it. Added a 15s timeout, matching `GITHUB_REF_RESOLUTION_TIMEOUT`.

- **Docs implying a namespace can map to `owner/repo`** — fair, the paragraph sits right under the GitHub section. Now says explicitly that a namespace maps to a hosted URL template and is an alternative to the GitHub form, not a shorthand for it.

- **Owner length boundary + empty item segments** — added. 39 chars resolves, 40 falls back, `owner/repo/` falls back. `owner/repo//button` resolves with the item name `/button`; I pinned that as-is with a comment rather than "fixing" it, since the item name is only ever matched against `registry.json` and never joined into a path, so it fails cleanly as an unknown item. Same as the React CLI.

- **Timeout branch untested** — added, with fake timers, asserting both the suggestion and that the signal was actually aborted.

**Not changed**

- **`isGitHubItemAddress` / `isGitHubRegistrySource` throwing.** Keeping the throw. Returning `false` for a malformed ref would send `owner/repo/button#my tag` on to the default registry, where it fails as an unknown item name and hides the actual mistake — a worse error for the user. This also matches `shadcn`, which I'd rather not diverge from here. I've documented the behaviour on both functions instead.

- **`HTTPS_PROXY` fallback and lazy agent construction in `proxy.ts`.** Both are genuine improvements, but `proxy.ts` is a straight lift of what `fetcher.ts` already did, and changing proxy resolution would change behaviour for every existing CLI user from inside a feature PR about GitHub registries. Happy to send it as a separate PR if you'd like it — just didn't want to smuggle it in here.

`pnpm test` green from the root (517 + 13). Also re-ran the live install against a real registry after the fetch changes — still resolves 39 files through a two-level `registryDependencies` chain.

PR #1927, opened 2026-08-06
feat: port shadcn/ui questionnaire component

zernonia:
Ports the [questionnaire](https://ui.shadcn.com/docs/components/base/questionnaire) component from shadcn/ui: a multi-step form that shows one question at a time with single-choice, multiple-choice, freeform, and skippable answers.

## What's included

- `registry/bases/reka/ui/questionnaire` (source of truth, `cn-*` placeholders) and the mirrored `registry/new-york-v4/ui/questionnaire`
- `cn-questionnaire-*` blocks in all 8 style CSS files, taken from the matching upstream style files
- 12 demos and a docs page at `/docs/components/questionnaire`
- Regenerated registry output (`pnpm registry:build`)

## Implementation notes

The React version builds on a headless primitive in `@shadcn/react` that has no Reka UI equivalent, so the navigation, answer registration, status, and validation logic lives in `useQuestionnaire.ts` next to the components. Two Vue-specific details are worth flagging for review:

- `QuestionnaireChoice.checked` defaults to `undefined`. Without it, Vue casts the absent boolean prop to `false` and every choice behaves as a controlled input that never updates.
- `QuestionnaireInput` tracks its own value. Vue re-applies `value` on every render (`patchElement` always patches the `value` key), so an `undefined` binding clears an uncontrolled input mid-typing.

React `render` props map to slots: `QuestionnaireProgress` exposes `current`, `total`, `first`, and `last`; `QuestionnaireChoice` exposes `checked`, `disabled`, `shortcut`, and `type`.

## Not ported

- The dialog example, which composes `QuestionnaireTitle` into `DialogTitle` through Base UI's `render` prop. The card example is included, with `QuestionnaireTitle` used directly inside `CardHeader` instead.
- The dev-only `console.warn` diagnostics that check `items` against the rendered items.

## Verification

`pnpm registry:build`, eslint, and `vue-tsc` are clean for the new files. All 12 demos were driven in a headless browser: progress and navigation, required/optional validation, skip, multiple selection, freeform answers, letter/number shortcuts, external validation with `invalid`, controlled navigation with `v-model:item`, conditional disabled items, native form reset restoring defaults, and arrow/Enter keyboard navigation. `FormData` submission was checked for each answer type, and no console errors were logged.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added a flexible, multi-step Questionnaire component with required and optional questions, single- and multiple-choice answers, text inputs, progress tracking, validation, reset, skip, and submit controls.
  * Added keyboard navigation, configurable shortcuts, focus management, accessibility support, and styled variants across supported themes.
* **Documentation**
  * Added comprehensive usage, composition, accessibility, keyboard navigation, and API documentation.
* **Examples**
  * Added interactive demos covering animated, card, conditional, controlled, dialog, freeform, navigation, validation, resume, skip, progress, multiple-choice, and shortcut-based questionnaires.


zernonia:
Pushed 4b71fe9a addressing the review, plus the `render`-prop gap.

## `as-child` composition (the Vue equivalent of `render`)

`QuestionnaireProgress`, `QuestionnaireTitle`, `QuestionnaireDescription`, `QuestionnaireChoices`, `QuestionnaireError`, `QuestionnaireActions`, and the four navigation buttons now render through Reka's `Primitive` and accept `as` / `as-child`. That unblocks the dialog example, which is now included as a demo:

```vue
<QuestionnaireTitle as-child>
  <DialogTitle>Which files are in scope?</DialogTitle>
</QuestionnaireTitle>
```

One wrinkle worth calling out: `QuestionnaireDescription` now adopts the id of the child it renders. Reka's `DialogContent` checks that its own `descriptionId` resolves to an element, so forcing the questionnaire's generated id onto `DialogDescription` broke the dialog's `aria-describedby` and logged a missing-description warning. The item and the dialog now describe the same element.

## Fixed

- **Re-sync the whole choice group after a controlled interaction** — checking a radio clears its siblings, so restoring only the changed input left the group with no checked control when the host kept the previous answer. `QuestionnaireItem` now exposes a sync tick that every choice watches.
- **`QuestionnaireInput` watches the value as well as `filled`** — a host update between two non-empty values now clears the skipped state, so the answer is not dropped from `FormData`.
- **`QuestionnaireInput` watches `props.disabled` when registering** — the registered `ownDisabled` could go stale while the item itself was disabled.

## Not changed, with reasons

- **`@lucide/vue` in the registry dependencies** — no component in the repo declares it. `checkbox` and `sonner` both import from `@lucide/vue` and declare only `reka-ui` / `vue-sonner`; the crawler's dependency allowlist in `scripts/crawl-content.ts` does not include it, and the CLI maps icons per base. Adding it here alone would be inconsistent.
- **The GitHub source link** — it points at `tree/dev/...`, which resolves once this merges. Same pattern as `marker.md` and the other component docs.
- **`aria-keyshortcuts` before selection** — the shortcut is already exposed. `getAnswerKeyShortcuts(shortcut, filled)` only makes the `Enter` half conditional, so an unchecked choice renders `aria-keyshortcuts="A"`; verified in the DOM.
- **`immediate: true` on the status watcher** — not needed here. Choices register their defaults during their own setup, which happens after the item's watcher is created, so an item that mounts already answered does emit `update:status`. Upstream's `onStatusChange` deliberately skips the mount call as well.
- **Freeform-specific error wording** — upstream uses one message for both cases, and this is a port.

## Verification

`pnpm registry:build`, eslint, and `vue-tsc` are clean for the new files. The three browser suites now cover 65 assertions across all 13 demos, including the new dialog composition: `as-child` title and description wiring, the dialog's accessible name and description, validation inside the dialog, and submit closing it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

zernonia:
Pushed 46f23625 for the second review pass.

## Fixed

**Items composed with `as-child` were unnamed.** `QuestionnaireTitle` now registers the id of whatever it renders, and the item labels itself with it whenever the title is not the fieldset legend. It happens automatically, so the composition stays a one-liner. The title keeps the child's own id rather than overriding it, which matters for children that manage their own id.

**Navigation ignored `disabled` on non-button render targets.** `QuestionnairePrevious`, `QuestionnaireSkip`, and `QuestionnaireNext` check `props.disabled` before navigating and call `preventDefault()`, and all four buttons expose `aria-disabled`. Verified with a throwaway page rendering each as `<a href="#next" disabled>`: clicking no longer advances the questionnaire or follows the anchor.

## The `as-child` example moved from the dialog to the card

Chasing the first finding turned up a worse problem underneath it. Reka's `DialogTitle` and `DialogDescription` both take their id from the dialog root, so rendering one per question produced **duplicate ids** — `getElementById` resolved to the first, and on question two the dialog and the fieldset were both announced with question one's title. Pinning explicit ids doesn't help either: that breaks the dialog's own `aria-labelledby`, which points at the id the dialog root generated. Upstream's dialog example has the same structure.

So the composition example is now the card, where each `CardTitle` / `CardDescription` gets a unique id and everything resolves correctly:

```vue
<QuestionnaireTitle as-child>
  <CardTitle>What should the agent work on?</CardTitle>
</QuestionnaireTitle>
```

The dialog demo keeps its place as "compose inside a Dialog with host-owned dismissal" — one `DialogTitle` and `DialogDescription` for the dialog, and each question keeps its own `legend`. The docs say why.

## Verification

Three browser suites, now 69 assertions. New ones cover: both card items named by their own title and described by their own description, the dialog named and described exactly once, questions keeping their legends, and a page-wide duplicate-id check that now returns empty. `registry:build`, eslint, and `vue-tsc` are clean.

One note: six unrelated docs files (`marker.md`, `attachment.md`, `message.md`, `message-scroller.md`, `shimmer.md`, `scroll-fade.md`) lost their `new: true` frontmatter in my working tree during this session. That is not my change, so I left it uncommitted.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1935, opened 2026-08-12
fix(docs): prevent component preview clipping

jevin98:
Fixes component demos whose content is taller than the default preview area being clipped before it reaches the code panel. Commit `03554c98` changed the shared preview wrapper in `ComponentPreviewTabs.vue` from `min-h-72` to the fixed height `h-72`; because the outer preview card uses `overflow-hidden`, taller demos such as Table lost their final rows, footer, and caption.

This restores `min-h-72`, preserving the 18rem default while allowing the preview wrapper to grow with its rendered component. The alignment changes from the original commit remain unchanged.

Table component preview clipped at the code panel boundary

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the component preview area to expand when its content exceeds the minimum height, preventing content from being constrained or clipped.

### DIFF
--- a/apps/v4/components/ComponentPreviewTabs.vue
+++ b/apps/v4/components/ComponentPreviewTabs.vue
@@ -35,7 +35,7 @@ const isMobileCodeVisible = ref(false)
           :data-align="align"
           :data-chromeless="chromeLessOnMobile"
           :class="cn(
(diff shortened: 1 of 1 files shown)


jevin98:
@zernonia I noticed that commit 03554c98 changed the component preview container from min-h-72 to the fixed height h-72. Could you share the reason for this adjustment?
Restoring min-h-72 fixes the clipping for taller demos, but I want to avoid affecting the behavior that the original change intended to support. If this revert impacts a specific scenario or design requirement, could you point it out? I’ll adjust the fix accordingly.

zernonia:
Thanks for the fix @jevin98 , I've gone through and fix those previewClass, should be all good now

